PR #3, opened 2021-02-18
CY-50 Adding Panel for Disk Statistics

babhalodia:
Adding panel which displays disk usage of Splunk Instances.

### DIFF
--- a/cyences_app_for_splunk/default/data/ui/views/cs_splunk_admin_reports.xml
+++ b/cyences_app_for_splunk/default/data/ui/views/cs_splunk_admin_reports.xml
@@ -16,15 +16,27 @@
     <latest></latest>
     <sampleRatio>1</sampleRatio>
   </search>
+  <search id="base_search">
+    <query>| rest splunk_server=* /services/server/status/partitions-space 
+  | eval free = if(isnotnull(available), available, free) 
+  | eval usage = round((capacity - free) / 1024, 2) 
+  | eval capacity = round(capacity / 1024, 2) 
+  | eval compare_usage = usage." / ".capacity 
+  | eval pct_usage = round(usage / capacity * 100, 2) 
+  | stats first(fs_type) as fs_type first(compare_usage) as compare_usage first(pct_usage) as pct_usage, by mount_point splunk_server 
+  | sort - pct_usage 
+  | table splunk_server mount_point,compare_usage,pct_usage 
+    </query>
+    <earliest>-24h@h</earliest>
+    <latest>now</latest>
+  </search>
   <fieldset submitButton="false"></fieldset>
   <row>
     <panel>
       <html>
+        <p>Splunk Admin Reports</p>
         <p>Shows information which is important for security and overall health of the Splunk environment.</p>
         <p>Alerts "Splunk Admin - Missing Data in the Index" and "Splunk Admin - Missing Forwarder" has to be enabled in order to populate this dashboard.</p>
-        <p>Both alerts update the information every hour.</p>
-        <p>Data on the Overview page might not match with the data in this dashboard if you have filtered any forwarders or indexes with macros cs_splunk_admin_missing_forwarders_filter or cs_splunk_admin_missing_indexes_filter. This dashboard shows all the indexes and forwarders.</p>
-        <p>To remove any forwarder or index from the list visit "Differentiating Between Missing Indexes &amp; Forwarders and Removed Indexes &amp; Forwarders" section from the <a href="https://cyences.com/cyences-app-for-splunk/" target="_blank">documentation</a>.</p>
       </html>
       <table>
         <search>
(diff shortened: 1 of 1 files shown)


cyences_app_for_splunk/default/data/ui/views/cs_splunk_admin_reports.xml
```diff
@@ -16,15 +16,27 @@
...
+    <latest>now</latest>
+  </search>
   <fieldset submitButton="false"></fieldset>
   <row>
     <panel>
       <html>
+        <p>Splunk Admin Reports</p>
         <p>Shows information which is important for security and overall health of the Splunk environment.</p>
         <p>Alerts "Splunk Admin - Missing Data in the Index" and "Splunk Admin - Missing Forwarder" has to be enabled in order to populate this dashboard.</p>
-        <p>Both alerts update the information every hour.</p>
```
VatsalJagani, line 25: Add these lines, looks like the Cerberos instance had old code.

cyences_app_for_splunk/default/data/ui/views/cs_splunk_admin_reports.xml
```diff
@@ -16,15 +16,27 @@
...
+  | table splunk_server mount_point,compare_usage,pct_usage 
+    </query>
+    <earliest>-24h@h</earliest>
+    <latest>now</latest>
+  </search>
   <fieldset submitButton="false"></fieldset>
   <row>
     <panel>
       <html>
+        <p>Splunk Admin Reports</p>
```
VatsalJagani, line 37: Remove this line, looks like the Cerberos instance had old code.

cyences_app_for_splunk/default/data/ui/views/cs_splunk_admin_reports.xml
```diff
@@ -16,15 +16,27 @@
     <latest></latest>
     <sampleRatio>1</sampleRatio>
   </search>
+  <search id="base_search">
+    <query>| rest splunk_server=* /services/server/status/partitions-space
```
VatsalJagani, line 20: Will the rest-handler will show the status of all the instances?

I would using index=introspection would cover all the full instances including HFs, IDXs, and other SHs.
Also, we will be able to calculate Disk, Memory, and CPU from a single query.

That way we can complete CY-50 and CY-51 together with just one search too.

cyences_app_for_splunk/default/data/ui/views/cs_splunk_admin_reports.xml
```diff
@@ -125,4 +137,39 @@
...
+        <choice value="&gt;80">&gt; 80</choice>
+        <choice value="&gt;60">&gt;60</choice>
+        <default>&gt;0</default>
+        <initialValue>&gt;0</initialValue>
+      </input>
+      <table>
+        <search base="base_search">
+          <query>| search splunk_server=$host$ AND pct_usage$disk_usage$| rename mount_point as "Mount Point", compare_usage as "Disk Usage (GB)", capacity as "Capacity (GB)", pct_usage as "Disk Usage (%)", splunk_server as Host | search Host=$host$</query>
+        </search>
+        <option name="drilldown">none</option>
```
VatsalJagani, line 166: I would say adding a drilldown would be helpful when some system is having the issue.
Drilldown: time chart of disk-usage over time. (may be for last 7 days would be useful.)

cyences_app_for_splunk/default/data/ui/views/cs_splunk_admin_reports.xml
```diff
@@ -16,15 +16,27 @@
     <latest></latest>
     <sampleRatio>1</sampleRatio>
   </search>
+  <search id="base_search">
```
VatsalJagani, line 19: Rename the id.
"base_search" is very generic, name something like "base_disk_search" or "base_usage_search".

cyences_app_for_splunk/default/data/ui/views/cs_splunk_admin_reports.xml
```diff
@@ -125,4 +137,39 @@
...
+        <choice value="*">All</choice>
+        <fieldForLabel>Host</fieldForLabel>
+        <fieldForValue>Host</fieldForValue>
+        <search base="base_search">
+          <query>| fields splunk_server | rename splunk_server as Host | dedup Host</query>
+        </search>
+        <default>*</default>
+        <initialValue>*</initialValue>
+      </input>
+      <input type="dropdown" token="disk_usage" searchWhenChanged="true">
```
VatsalJagani, line 154: Having token names prefixed with tkn_ is good practice.

VatsalJagani (changes requested):

cyences_app_for_splunk/default/data/ui/views/cs_splunk_admin_reports.xml
babhalodia: Rest will provide all the Splunk instances which include Search Heads, Indexers, CM, DS but not HFs. I am changing the search to use _introspection index.

cyences_app_for_splunk/default/data/ui/views/cs_splunk_admin_reports.xml
babhalodia: Okay. I will add it.

VatsalJagani:
Closing PR as already covered in - https://github.com/VatsalJagani/Splunk-Cyences-App-for-Splunk/pull/88/files

PR #7, opened 2021-04-12
CY-181 - Adding & Updating Filters for Globally Detected Malicious IPs Panel

babhalodia:
Adding & Updating Filters For "Globally Detected Malicious IPs" Dashboard

### DIFF
--- a/cyences_app_for_splunk/default/data/ui/views/cs_malicious_ip_list.xml
+++ b/cyences_app_for_splunk/default/data/ui/views/cs_malicious_ip_list.xml
@@ -31,7 +31,8 @@
           </ul>
         </p>
         <p>The Malicious IP List contains IP addresses discovered in the last seven days by default. The time range can be changed by updating the following macro definition (Settings &gt; Configuration): cs_palo_malicious_ip_list_filter_old_results.</p>
-        <p><b>Data Collection:</b> The pan:traffic sourcetype for Palo Alto is being used to collect data via the Palo Alto Networks Add-on for Splunk.</p>
+        <p>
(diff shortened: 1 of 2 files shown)


cyences_app_for_splunk/default/data/ui/views/cs_malicious_ip_list.xml
```diff
@@ -69,24 +70,42 @@
...
         <label>Search for Bad IP Addresses</label>
+        <default>*</default>
+        <suffix>*</suffix>
+      </input>
+      <input type="dropdown" token="tkn_location">
+        <label>Search By Location</label>
+        <search base="search_list">
+          <query> | search ip=$tkn_ip|s$ description=$tkn_description|s$  |stats count by ip_location</query>
+        </search>
         <choice value="*">All</choice>
```
VatsalJagani, line 83: Bro,  in some cases IP location will not be present, so those will not be listed because of condition `iplocation=*` as default. We need to implement something like this.

Ex.
`<input type="text" token="tkn_ip_tmp">
        <label>IP Address</label>
        <default></default>
        <change>
          <condition match="'tkn_ip_tmp'=&quot;&quot;">
            <set token="tkn_ip">()</set>
          </condition>
          <condition>
            <set token="tkn_ip">(ip="*$tkn_ip_tmp$*")</set>
          </condition>
        </change>
      </input>`
      
      See Device Master Table dashboard for reference. See input, init at the top of the dashboard, and tokens in the search queries.

VatsalJagani (changes requested):

VatsalJagani (approved):

PR #10, opened 2021-04-22
CY-206-Enhancement-Antivirus-Panel

babhalodia:


cyences_app_for_splunk/default/data/ui/views/cs_overview.xml
```diff
@@ -226,7 +226,33 @@
       <title>Antivirus</title>
       <table>
         <search base="basesearch">
-          <query>| `cs_filter_savedsearches("Sophos|Windows Defender|CrowdStrike")`</query>
+          <query>| `cs_filter_savedsearches("Sophos|Windows Defender|CrowdStrike")` 
+| eval Name=if(Name="Sophos Service is not Running","Sophos Service Stopped/Disabled Events",Name)
```
VatsalJagani, line 230: case would be a better option.

cyences_app_for_splunk/default/data/ui/views/cs_windows_defender_reports.xml
```diff
@@ -68,8 +68,12 @@
       </input>
       <table>
         <search>
-          <query>`cs_windows_defender` EventCode=1151 | dedup host | search $tkn_host$ $tkn_signature_version$ 
-| table _time, host, RTP_state, Platform_version, Engine_version, AVSignature_version, AV_signature_creation_time, BM_state, IOAV_state, OA_state, Last_full_scan_start_time, Last_full_scan_end_time, Last_quick_scan_start_time, Last_quick_scan_end_time
+          <query>`cs_windows_defender` EventCode=1151
+| search $tkn_host$ $tkn_signature_version$ 
+| dedup host 
+| rex field=_raw "&lt;Data Name='RTP\s+state'&gt;(?&lt;RTP_state_cs&gt;[^\&lt;]*)"
```
VatsalJagani, line 74: Perfect.

cyences_app_for_splunk/default/data/ui/views/cs_overview.xml
```diff
@@ -226,7 +226,33 @@
...
+    | fields host dhost type 
+    | stats latest(type) as type by host, dhost 
+    | where type=="Event::Endpoint::SavDisabled" 
+    | stats count as Count 
+    | eval Name="Currently Stopped/Disabled Sophos RealTime Protection"]
+| append
+    [| search `cs_wineventlog_system` SourceName="Microsoft-Windows-Service Control Manager" Message="The Windows Defender Antivirus Service service entered the *"  earliest=$timeRange.earliest$ latest=$timeRange.latest$ | stats latest(Message) as Message by host | search Message=*stopped* | stats count as Count
+    | eval Name="Currently Disabled/Stopped Windows Defender Services"]
+| append
+    [| search `cs_windows_defender` EventCode=1151  earliest=$timeRange.earliest$ latest=$timeRange.latest$ | fields RTP_state host| stats latest(RTP_state) as RTP_state by host | where RTP_state=="Disabled" | stats count as Count
```
VatsalJagani, line 253: I think each defender sends EventCode=1151 events every hour. So we can reduce the timerange to like last 4 hours only. We might need to do the same for the Health Report panel on the defender dashboard.
- But first, we need to confirm the fact that these events are being sent by all the hosts at regular intervals.

VatsalJagani (approved):

cyences_app_for_splunk/default/data/ui/views/cs_overview.xml
VatsalJagani: use macro.

babhalodia: This will not work. I checked the logs and for every 4 hours of span, we are not getting the same number of hosts. So, currently, we are not implementing this.

PR #23, opened 2021-06-01
CY-218 Whitelisting version for rdpinit.exe + rdpshell.exe

babhalodia:


cyences_app_for_splunk/default/macros.conf
```diff
@@ -119,7 +119,7 @@ iseval = 0
...
+definition = search NOT process_path IN ("C:\\Program Files\\Common Files\\microsoft shared\\MSInfo\\msinfo32.exe","C:\\Users\\*\\AppData\\Local\\Temp\\*\\DismHost.exe","C:\\Users\\*\\AppData\\Local\\Microsoft\\OneDrive\\Update\\OneDriveSetup.exe","C:\\Users\\*\\AppData\\Local\\Microsoft\\OneDrive\\StandaloneUpdater\\OneDriveSetup.exe","C:\\Windows\\regedit.exe","C:\\Windows\\hh.exe","C:\\Users\\*\\AppData\\Local\\Temp\\WindowsUpdateLog\\tracerpt.exe","C:\\Program Files\\Microsoft RDInfra\\StackSxS\\1.0.*.*\\rdpinit.exe","C:\\Program Files\\Microsoft RDInfra\\StackSxS\\1.0.*.*\\rdpshell.exe","C:\\Windows\\Temp\\*\\DismHost.exe")
```
VatsalJagani, line 122: Bro, we can actually put the whole version as *
And that is safe to do, as most of the ransomware skips program files while encrypting files.

babhalodia: Whitelisted all the version.

PR #41, opened 2021-07-19
CY-257: Added support for tenable:sc.

VatsalJagani:


cyences_app_for_splunk/default/props.conf
```diff
@@ -299,3 +299,54 @@ EVAL-user = coalesce(User, user, "-")
...
 EVAL-LogName = if(isnull(LogName), "System", LogName)
 EVAL-user = coalesce(User, user, "-")
+
+
+
+######################
+#### Tenable #########
+######################
+[tenable:sc:assets]
+EVAL-tenable_uuid = if(isnull(uuid) or trim(uuid)="", lower(ip), replace(uuid, " ", "*"))
```
babhalodia, line 309: In tenable:sc:assets events, we are getting hostUniqueness field. So instead of just using ip as tenable_uuid, we should use combination of reposioryID,ip,dnsname as unique host.

babhalodia (changes requested):

babhalodia (approved):

PR #42, opened 2021-07-19
CY-242: Showing High vulnerabilities count.

VatsalJagani:


cyences_app_for_splunk/default/macros.conf
```diff
@@ -156,8 +156,9 @@ definition = inputlookup cs_device_inventory \
 | eval tmp_crowdstrike_status = crowdstrike_status, crowdstrike=case(tmp_crowdstrike_status == "Present", 0) \
 | fieldformat crowdstrike=tmp_crowdstrike_status \
 | eval ip_drilldown=mvjoin(ip, ","), hostname_drilldown=mvjoin(hostname, ","), user_drilldown=mvjoin(user, ",") \
-| eval active_vul=coalesce(active_vuln, ACTIVE), total_vul=coalesce(total_vuln, TOTAL_VULNS) \
-| table _time, uuid, ip, hostname, mac_address, user, os, lansweeper, qualys, tenable, active_vul, total_vul, sophos, defender, crowdstrike, tmp_*, *_last_event, ip_drilldown, hostname_drilldown, user_drilldown \
+| eval active_vul=coalesce(active_vuln,ACTIVE), total_vul=coalesce(total_vuln,TOTAL_VULNS), severity_5_q=if(isnull(ACTIVE_SEVERITY_5), 0, ACTIVE_SEVERITY_5), severity_4_q=if(isnull(ACTIVE_SEVERITY_4), 0, ACTIVE_SEVERITY_4), high_vul_q=severity_5_q+severity_4_q, high_vul_t=critical_active_vuln+high_active_vuln, high_vul_q=if(isnull(high_vul_q), 0, high_vul_q), high_vul_t=if(isnull(high_vul_t), 0, high_vul_t), high_vul=high_vul_q+high_vul_t \
```
babhalodia, line 159: Before adding, first check whether values are null or not in below eval:
high_vul_q=severity_5_q+severity_4_q

Checking after adding will not work because it is possible that severity_5_q is null but other one is not. But we will get null as result. 


If severity_5_q & severity_4_q will never null then no need to check for below:
high_vul_q=if(isnull(high_vul_q), 0, high_vul_q)

babhalodia (approved):

babhalodia (approved):

PR #43, opened 2021-07-20
CY-248: Lansweeper dashboard enhancement.

VatsalJagani:


cyences_app_for_splunk/default/data/ui/views/cs_lansweeper.xml
```diff
@@ -417,6 +456,114 @@
...
+      <input type="dropdown" token="tkn_tenable_logs_mac">
+        <label>Tenable Data</label>
+        <choice value="*">All</choice>
+        <choice value="Present">Present</choice>
+        <choice value="Absent">Absent</choice>
+        <default>*</default>
+      </input>
+      <table>
+        <search base="search_mac_assets">
+          <query>| search $tkn_host$ $tkn_ip$ $tkn_user$ $tkn_asset_state$ Qualys=$tkn_qualys_logs_linux|s$ Tenable=$tkn_tenable_logs_mac|s$</query>
```
babhalodia, line 478: I think there is some mistake over here: Qualys=$tkn_qualys_logs_linux|s$. Instead of "Linux" in the token it should be "Mac"

babhalodia (approved):

PR #52, opened 2021-08-02
Cy 225 microsoft graph security score

babhalodia:


cyences_app_for_splunk/default/data/ui/views/cs_o365_reports.xml
```diff
@@ -53,6 +60,60 @@
       </table>
     </panel>
   </row>
+  <row>
+    <panel>
+      <title>Azure - Current Security Score(In Percentage)</title>
+      <single>
+        <search>
+          <query>`cs_azure_securityscore` sourcetype="graphsecurity:score"  $tkn_host$ | eval percentage = (currentScore/maxScore)*100 | timechart max(percentage) as percentage | streamstats last(percentage) as percentage</query>
```
VatsalJagani, line 68: Bro, we should use latest(percentage) instead of max. As we are looking for latest value of security score, there is chance of security score going down.

cyences_app_for_splunk/default/data/ui/views/cs_o365_reports.xml
```diff
@@ -53,6 +60,60 @@
...
+          <earliest>-7d@h</earliest>
+          <latest>now</latest>
+        </search>
+        <option name="numberPrecision">0.0</option>
+        <option name="rangeColors">["0x53a051","0x0877a6","0xf8be34","0xf1813f","0xdc4e41"]</option>
```
VatsalJagani, line 73: Make sure coloring scheme is what you intended.

cyences_app_for_splunk/default/data/ui/views/cs_o365_reports.xml
```diff
@@ -53,6 +60,60 @@
...
+        <option name="showTrendIndicator">1</option>
+        <option name="unit">%</option>
+        <option name="drilldown">all</option>
+        <drilldown>
+          <set token="tkn_panel_basis_security_score_visible">"True"</set>
```
VatsalJagani, line 78: Shall we add note in html that this panel has drilldown?

cyences_app_for_splunk/default/data/ui/views/cs_o365_reports.xml
```diff
@@ -53,6 +60,60 @@
...
+        </drilldown>
+      </single>
+    </panel>
+  </row>
+  <row depends="$tkn_panel_basis_security_score_visible$">
+    <panel>
+      <title>Parameter used to calculate Security Score</title>
+      <table id="tbl_security_score">
+        <search>
+          <query>`cs_azure_securityscore` sourcetype="graphsecurity:score" | where $tkn_panel_basis_security_score_visible$=="True"
```
VatsalJagani, line 104: may ne set token to empty string (instead of True) and use token at the beginning of the search so search won't execute until usrr drilldown.

cyences_app_for_splunk/default/macros.conf
```diff
@@ -489,6 +489,10 @@ iseval = 0
 definition = search *
 iseval = 0
 
+[cs_azure_securityscore]
+definition = index=*
```
VatsalJagani, line 493: Bro, lets use o365 index. That way we'll avoid index=*. Also, please Cyences doc change that I did to make sure I added correct information about this.

VatsalJagani (changes requested):

cyences_app_for_splunk/default/macros.conf
babhalodia: bro we used SourceType in the search query so it will not use much resource. But I will change the index.

cyences_app_for_splunk/default/data/ui/views/cs_o365_reports.xml
babhalodia: Bro Search will not execute in this query also until user clicked.

cyences_app_for_splunk/default/data/ui/views/cs_o365_reports.xml
babhalodia: Added

cyences_app_for_splunk/default/data/ui/views/cs_o365_reports.xml
VatsalJagani: yeah i see it.

VatsalJagani (approved):

PR #55, opened 2021-08-05
Nc 92 monitor group membership changes

babhalodia:


cyences_app_for_splunk/default/savedsearches.conf
```diff
@@ -456,6 +456,57 @@ search = `cs_ad_active_directory` "objectClass=top|group" ("admonEventType=Sync"
 | eval _key=key_val \
 | outputlookup cs_ad_obj_group append=true
 
+[AD - cs_ad_obj_computer - Lookup Gen]
```
VatsalJagani, line 459: Bro, kindly update the Reports list macro.

cyences_app_for_splunk/default/savedsearches.conf
```diff
@@ -456,6 +456,57 @@ search = `cs_ad_active_directory` "objectClass=top|group" ("admonEventType=Sync"
...
+disabled = 0
+enableSched = 0
```
VatsalJagani, line 461: You mean 1 right? We need to keep scheduling enabled.

cyences_app_for_splunk/default/savedsearches.conf
```diff
@@ -456,6 +456,57 @@ search = `cs_ad_active_directory` "objectClass=top|group" ("admonEventType=Sync"
...
+action.email.useNSSubject = 1
```
VatsalJagani, line 462: We can remove this parameter is not used.

cyences_app_for_splunk/default/savedsearches.conf
VatsalJagani: I think we need to check this parameter for other lookup gen report of AD.

VatsalJagani (changes requested):

PR #56, opened 2021-08-05
Fixing the issue with active directory related alerts.

VatsalJagani:
Bro, let's complete this if you don't mind.


VatsalJagani:
Bro, one more thing, it seems like duplicate panels for the same purpose on the "Active Directory and Windows" dashboard. I think we need to look into that.

babhalodia (approved):

PR #110, opened 2021-12-02
Cy 310 add panel to remove host from lookup

babhalodia:


cyences_app_for_splunk/default/data/ui/views/cs_decommision_host.xml
```diff
@@ -0,0 +1,95 @@
...
+      </table>
+      <html>
+        <p>Note: Query will remove the Host from the Linux User List Lookup and you will not see the host in the dashboard panel called "User List".</p>
+      </html>
+    </panel>
+  </row>
+  <row>
+    <panel>
+      <title>Decommission Windows Host from Windows Lookup</title>
+      <input type="dropdown" token="tkn_remove_forwarder_dropdown">
```
VatsalJagani, line 52: for all the tokens, we should add linux or windows postfix. Currently they are conflicting.

VatsalJagani (changes requested):

VatsalJagani (approved):

PR #146, opened 2022-04-12
Reduce the False postive result for the antivirus alerts using lansweeper asset

mahirchavda:
Reduce the False postive result for the antivirus alerts using lansweeper asset


VatsalJagani (approved):
Please make sure to update the release notes in a separate open PR for these changes.

PR #159, opened 2022-05-09
CY-350: Add private ip details in the vpn dashboad

mahirchavda:
* Add Successful Session panel to show private ip detail
* Fix the drilldown query in the Login Details panel

**Note**: The private ip field is not present in the sourcetype=pan:system events (Legacy Palo Alto events)

### DIFF
--- a/cyences_app_for_splunk/default/data/ui/views/cs_vpn_reports.xml
+++ b/cyences_app_for_splunk/default/data/ui/views/cs_vpn_reports.xml
@@ -197,9 +197,40 @@
         </search>
         <option name="count">5</option>
         <option name="refresh.display">progressbar</option>
+        <option name="drilldown">row</option>
         <option name="rowNumbers">true</option>
         <drilldown>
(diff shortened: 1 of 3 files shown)


VatsalJagani (commented):
- Performance difference between tstats query vs regular query.
- tstats query is being used elsewhere for VPN-related stuff as well. Just grep with index macro "cs_vpn_indexes" to see if we are missing anything.
- Please also check tags.conf and eventtypes.conf if we need any changes in that.

cyences_app_for_splunk/default/data/ui/views/cs_vpn_reports.xml
```diff
@@ -197,9 +197,40 @@
...
+      </table>
+    </panel>
+  </row>
+  <row>
+    <panel>
+      <title>Successful Session</title>
+      <table>
+        <search>
+          <query>`cs_vpn_indexes` dest_category="vpn_auth" action="success" $User$
+| stats values(src) as src_ip, values(private_ip) as private_ip by action, user, _time
```
VatsalJagani, line 214: Why don't we just use table? as we are doing group by _time, stats is not required.

VatsalJagani (approved):

PR #178, opened 2022-05-19
CY-382: Add AD - Bulk User Creation or Deletion alert & filter macro

mahirchavda:
Add AD - Bulk User Creation or Deletion alert & filter macro

### DIFF
--- a/cyences_app_for_splunk/default/data/ui/views/cs_configuration.xml
+++ b/cyences_app_for_splunk/default/data/ui/views/cs_configuration.xml
@@ -531,6 +531,24 @@
             </th>
           </tr>
           <tr>
+            <th style="text-align:left">cs_ad_bulk_user_creation_deletion_count_limit</th>
+            <th style="text-align:left">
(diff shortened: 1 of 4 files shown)


cyences_app_for_splunk/default/data/ui/views/cs_configuration.xml
```diff
@@ -530,6 +530,24 @@
               <button id="macro_cs_ad_password_change_outside_working_hour_filter_button">Update</button>
             </th>
           </tr>
+          <tr>
+            <th style="text-align:left">cs_ad_bulk_user_creation_deletion_count_limit</th>
```
VatsalJagani, line 534: We are not generally adding limit related macros with filter categories. So let's remove it for now, we'll see if we need it, we'll add it in the right place.

VatsalJagani (approved):

cyences_app_for_splunk/default/data/ui/views/cs_configuration.xml
mahirchavda: As discussed over chat, keeping as it is as we already have few limit macros in the configuration dashboard

PR #185, opened 2022-06-07
Use the Splunk map for the network report dashboard

mahirchavda:


VatsalJagani (changes requested):
1. Change from bytes to count
2. Use camel-case labels as already provided in the description (Blue, Red, instead of blue, red)

PR #207, opened 2022-08-17
Fix the late alert time issue for the forensic dashboard

mahirchavda:


cyences_app_for_splunk/default/alert_actions.conf
```diff
@@ -0,0 +1,4 @@
+[cyences_alert_action]
+label = Write to Cyences Index
```
VatsalJagani, line 2: label = Cyences alert action

cyences_app_for_splunk/default/alert_actions.conf
```diff
@@ -0,0 +1,4 @@
...
+description = Write to Cyences Index
```
VatsalJagani, line 3: description = Write alert results into cyences index

VatsalJagani (changes requested):
Please also remove the relevant instruction (making manual props.conf changes) from the document.

VatsalJagani (approved):

PR #215, opened 2022-11-01
Add firewall host information in the dashboard for palo events

mahirchavda:
dest is a eval field in the Authentication datamodel. If not present, default value will be unknown


VatsalJagani (commented):
Please check if any alerts can be improved as well.

PR #221, opened 2022-11-24
Code refactor

VatsalJagani:


cyences_app_for_splunk/bin/cyences_email_utility.py
```diff
@@ -263,12 +233,41 @@ def _get_smtp_creds_util(self):
...
+                self.logger.error("Not all of the recipient email domains are on the allowed domain list. Sending emails only to %s" % str(validRecipients))
+            
+            return validRecipients
+
+        else:
+            return recipients
 
 
-    def send(self, to, cc='', bcc='', subject='Splunk Alert', html_body='', results_link=''):
+    def send(self, to, cc=[], bcc=[], subject='Splunk Alert', html_body=''):
```
mahirchavda, line 266: If we use object as a default argument (list, or dict), it will keep reference of previously called item instead of passing the empty list.

cyences_app_for_splunk/bin/cyences_send_digest_email.py
```diff
@@ -13,113 +12,168 @@
...
-                continue
+            if 'cyences_severity' in event:
+                severity = event.get('cyences_severity', '').lower()
+
+                # Skip the event if cyences_severity is defined and not present in the severity_filter
+                if severity not in severity_filter:
+                    continue
+
+                try:
+                    event['__cyences_severity'] = ALERT_SEVERITIES[event['cyences_severity']]  # Add addition information for sorting
```
mahirchavda, line 63: should we use `severity` instead of `event['cyences_severity']` to make it case insensitive.

mahirchavda (approved):

PR #236, opened 2022-12-19
Fix action field for cisco:estreamer logs.

VatsalJagani:


cyences_app_for_splunk/default/props.conf
```diff
@@ -174,7 +174,7 @@ EVAL-action = "failure"
 FIELDALIAS-1vendor_user = user AS vendor_user
 EVAL-user = username
 EVAL-dest_category = if(rec_type_desc="New VPN Device Login", "vpn_auth", dest_category)
-EVAL-action = if(rec_type_desc="New VPN Device Login", "success", action)
+EVAL-action = if(rec_type_desc="New VPN Device Login", "success", case(fw_rule_action="Allow","allowed", like(fw_rule_action,"%Block%"),"blocked",1=1,action))
```
mahirchavda, line 177: Can we replace if with case to simply it little bit.

mahirchavda (approved):

PR #259, opened 2023-01-24
CY-453: Add dest and ip filter, make user multiselect filter

mahirchavda:


cyences_app_for_splunk/default/data/ui/views/cs_vpn_reports.xml
```diff
@@ -32,6 +32,18 @@
         <latest>now</latest>
       </default>
     </input>
+    <input type="dropdown" token="tkn_dest" searchWhenChanged="true">
+      <label>Dest</label>
```
VatsalJagani, line 36: Let's title it as Host.

VatsalJagani (approved):

cyences_app_for_splunk/default/data/ui/views/cs_vpn_reports.xml
mahirchavda, line 36:
```suggestion
      <label>Host</label>
```

PR #260, opened 2023-01-24
export alerts to system

mahirchavda:


TA-cyences/metadata/default.meta
```diff
@@ -31,6 +31,15 @@ export = system
 export = system
 
 
+### ALERT ACTIONS
+
+[alert_actions]
+export = system
+
+[alerts]
```
VatsalJagani, line 39: Why alerts?

TA-cyences/metadata/default.meta
```diff
@@ -31,6 +31,15 @@ export = system
 export = system
 
 
+### ALERT ACTIONS
+
+[alert_actions]
+export = system
```
VatsalJagani, line 37: Have we tested below:
When we open Cyences' alert from other App and make changes to alert action parameters, will it update the parameter under etc/apps/cyences_app_for_splunk/local/savedsearches.conf?

TA-cyences/metadata/default.meta
mahirchavda: Alert_actions to show the alerts action and alerts to show the alert form/html. (without alerts it won't show any fields under the alert action)

TA-cyences/metadata/default.meta
mahirchavda: yes. it is working as expected

VatsalJagani (approved):

PR #263, opened 2023-01-25
Bug fix linux success logins

babhalodia:


VatsalJagani (commented):
Bro, please ensure it works in other environment as it's doing today and make sure count still shows accurate.

babhalodia:
Checked with Etihad also.
And there also same issue.

PR #264, opened 2023-01-25
Show count with user and source field in the bruteforce alerts

mahirchavda:


VatsalJagani (approved):
Have we checked the Authentication dashboard which shows similar data?

mahirchavda:
> Have we checked the Authentication dashboard which shows similar data?

Yes. Not applicable to dashboard as it shows different data

PR #267, opened 2023-02-01
Add probable source in the ad password change alert and change severity to medium

mahirchavda:
…

### DIFF
--- a/cyences_app_for_splunk/bin/cy_timestamp_filter.py
+++ b/cyences_app_for_splunk/bin/cy_timestamp_filter.py
@@ -1,0 +1,31 @@
+import sys
+
+from splunklib.searchcommands import dispatch, StreamingCommand, Configuration, Option, validators
+
+import logging
+import logger_manager
+
+logger = logger_manager.setup_logging("cy_timestamp_filter", logging.INFO)
+
+@Configuration(local=True)
+class CYTimestampFilter(StreamingCommand):
+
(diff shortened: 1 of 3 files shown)


VatsalJagani (changes requested):

VatsalJagani (approved):
We may further enhance this to use with many more alerts including maybe WinEventLog cleared. This sounds very promising custom command.

But I just realized, this could have been done by mvmap function of eval. But python seems much simpler.

mahirchavda:
> We may further enhance this to use with many more alerts including maybe WinEventLog cleared. This sounds very promising custom command.
> 
> But I just realized, this could have been done by mvmap function of eval. But python seems much simpler.

just curious; can we? we need to first apply time condition if it matches then we need to select same index number value but from another field. Also if combine then, we need to separate it in the mvmap so not sure.

VatsalJagani:
> > We may further enhance this to use with many more alerts including maybe WinEventLog cleared. This sounds very promising custom command.
> > But I just realized, this could have been done by mvmap function of eval. But python seems much simpler.
> 
> just curious; can we? we need to first apply time condition if it matches then we need to select same index number value but from another field. Also if combine then, we need to separate it in the mvmap so not sure.

Here you go:
`| eval output=mvmap(input, if(tonumber(mvindex(split(input, "###"), 0))>(_time-300) AND tonumber(mvindex(split(input, "###"), 0))<(_time+300), mvindex(split(input, "###"), 1), null()))`

PR #277, opened 2023-02-27
Bump addon and app version

mahirchavda:


TA-cyences/ReadMe.md
```diff
@@ -42,6 +42,9 @@ To uninstall the app, the users can follow the below steps:
 
 RELEASE NOTES
 -------------
+Version 1.1.0 (March 2023)
+* Added users.sh and groups.sh scripted inputs.
```
babhalodia, line 46: Added users.sh and groups.sh scripted inputs.  And added alerts which will trigger when there is any change for User or Group.

PR #290, opened 2023-03-22
Alert filter as multiselect in the SOC dashboard

mahirchavda:


cyences_app_for_splunk/default/data/ui/views/cs_soc.xml
```diff
@@ -27,7 +27,7 @@
         <latest>now</latest>
       </default>
     </input>
-    <input type="dropdown" token="tkn_savedsearch">
+    <input type="multiselect" token="tkn_savedsearch" searchWhenChanged="true">
```
VatsalJagani, line 30: Approved considering searches will be handled properly with the same token value.

VatsalJagani (approved):

PR #327, opened 2023-06-13
CY-497: Fix the permission issue for notable event assignment for normal user role

mahirchavda:


TA-cyences/metadata/default.meta
```diff
@@ -45,3 +45,7 @@ export = system
 [viewstates]
 access = read : [ * ], write : [ * ]
 export = system
+
+
```
VatsalJagani, line 49: Kindly add why we added this in the comment.

# Permission updated so all users can update the notable event status and assignee on Splunk.

VatsalJagani (approved):

PR #354, opened 2023-07-31
Added the alert for windows certificate expiration

hardikhdholariya:
- Added the alert which captures certificates that will expire within 7 days, and if they are expiring within one day, the severity will be critical, otherwise, the severity will be high.


VatsalJagani (approved):

mahirchavda:
I would suggest to not use critical severity. keep high and medium. @VatsalJagani What do you think?

VatsalJagani:
> I would suggest not to use critical severity. keep high and medium. @VatsalJagani What do you think?

Usama suggested critical for any actionable item.

mahirchavda (approved):

PR #358, opened 2023-08-08
Migrated the storage logic from file to kvstore

hardikhdholariya:
- Migrated the storage logic from file to kvstore
- Updated addentries, getdevices, cleanup operations
NOTE: merge operation is not updated.

### DIFF
--- a/cyences_app_for_splunk/bin/device_inventory_v2_util.py
+++ b/cyences_app_for_splunk/bin/device_inventory_v2_util.py
@@ -127,26 +127,26 @@ class DeviceManager:
             device_uuid = dm.add_device_entry(new_device_entry)
     """
 
-    def __init__(self, session_key, logger, operation, hostname_postfixes=[]):
+    def __init__(self, session_key, logger, hostname_postfixes=[]):
         self.session_key = session_key
         self.logger = logger
-        self.operation = operation
         self.hostname_postfixes = [
             element.strip() for element in hostname_postfixes if element.strip()
         ]
-        self.modify_devices = []
+        self.updated_devices = []
+        self.deleted_devices = []
         self.devices = self.read_kvstore_lookup(DEVICE_INVENTORY_LOOKUP_COLLECTION)
-        if self.devices:
-            self._convert_str_to_dict()
 
-    def _convert_str_to_dict(self):
-        for dvc in self.devices:
+    def _convert_str_to_dict(self, lookup_data):
+        lookup_data = lookup_data if lookup_data else []
+        for dvc in lookup_data:
             dvc["product_info"] = json.loads(dvc["product_info"])
             dvc["product_names"] = json.loads(dvc["product_names"])
             dvc["product_uuids"] = json.loads(dvc["product_uuids"])
             dvc["ips"] = json.loads(dvc["ips"])
             dvc["mac_addresses"] = json.loads(dvc["mac_addresses"])
             dvc["hostnames"] = json.loads(dvc["hostnames"])
+        return lookup_data
 
     def read_kvstore_lookup(self, collection_name):
         _, serverContent = rest.simpleRequest(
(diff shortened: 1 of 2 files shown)


cyences_app_for_splunk/bin/device_inventory_v2_util.py
```diff
@@ -13,34 +12,100 @@
...
+import json
+import cs_utils
+
+from splunk import rest
 
 
-MAX_TIME_EPOCH = 2147483647   # Tue Jan 19 2038 03:14:07
+MAX_TIME_EPOCH = 2147483647  # Tue Jan 19 2038 03:14:07
+
+CY_HOSTNAME_POSTFIXES_MACRO = "cs_device_inventory_hostname_postfixes"
```
VatsalJagani, line 27: I don't think this is required here as we are passing it as class argument.

cyences_app_for_splunk/bin/device_inventory_v2_util.py
```diff
@@ -49,349 +114,425 @@ def _internal_check_list_field_format(self, field):
...
+        self.session_key = session_key
+        self.logger = logger
+        self.operation = operation
+        self.hostname_postfixes = [
+            element.strip() for element in hostname_postfixes if element.strip()
+        ]
+        self.modify_devices = []
+        self.devices = self.read_kvstore_lookup(DEVICE_INVENTORY_LOOKUP_COLLECTION)
+        if self.devices:
+            self._convert_str_to_dict()
```
VatsalJagani, line 140: I would suggest keeping this inside the read_kvstore_lookup function itself.
OR
use self.read_devices(<lookup>)
            self._read_kvstore_lookup(<lookup>)
            and _convert here

VatsalJagani: like update_kvstore_entries is handling it internally.

cyences_app_for_splunk/bin/device_inventory_v2_util.py
```diff
@@ -49,349 +114,425 @@ def _internal_check_list_field_format(self, field):
...
+                sessionKey=self.session_key,
+                raiseAllErrors=True,
+            )
 
+    def __enter__(self):
+        return self
 
-    def is_match(self, device_entry, hostname_postfixes=[]):
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        if self.operation == "cleanup":
```
VatsalJagani, line 224: Keep two list:
* updated_devices
* deleted_devices

Perform operation on both lists on all the exits.
That way we don't need self.operation

VatsalJagani: Before updated_devices execution, remove entries from the updated_devices list that are also part of deleted_devices.

cyences_app_for_splunk/bin/device_inventory_v2_util.py
VatsalJagani: Keep the self.devices up to date.
* Update self.devices when updated. And add the key to modified_device_list. (While storing build the dict by using self.devices) -> For memory optimization and avoid conflicts
* Remove self.devices when deleted. And add the key to deleted_device_list.

VatsalJagani (changes requested):

cyences_app_for_splunk/bin/device_inventory_v2_util.py
hardikhdholariya: We are passing the values of the hostname_postfixes from this macro.

cyences_app_for_splunk/bin/device_inventory_v2_util.py
hardikhdholariya: Done

cyences_app_for_splunk/bin/device_inventory_v2_util.py
hardikhdholariya: Done

cyences_app_for_splunk/bin/device_inventory_v2_util.py
VatsalJagani: Yes, but we are getting it as an argument. It is defined in two python files.

hardikhdholariya: It's my bad. removed unused variable.

VatsalJagani (approved):

PR #361, opened 2023-08-10
Updated the sophos alerts

hardikhdholariya:
- Updated the severity of "Sophos RealTime Protection Disabled" to critical if it stayed disabled for more than 3 hrs.
- Updated the severity of "Failed to clean up threat by Sophos" alert to critical.

### DIFF
--- a/cyences_app_for_splunk/default/props.conf
+++ b/cyences_app_for_splunk/default/props.conf
@@ -553,7 +553,6 @@ EVAL-dest_name = mvappend(DNS, NETBIOS)
 # For Cyences Assets datamodel
 FIELDALIAS-asset_id = HOST_ID AS asset_id
 FIELDALIAS-asset_ip = IP AS asset_ip
-LOOKUP-qualys_kb_lookup = qualys_kb_lookup QID OUTPUT TITLE as description
 EVAL-asset_hostname = mvjoin(mvappend(DNS, NETBIOS), "~~")
 EVAL-state = "Active"
 
--- a/cyences_app_for_splunk/default/props.conf
+++ b/cyences_app_for_splunk/default/props.conf
@@ -553,7 +553,7 @@ EVAL-dest_name = mvappend(DNS, NETBIOS)
 # For Cyences Assets datamodel
 FIELDALIAS-asset_id = HOST_ID AS asset_id
 FIELDALIAS-asset_ip = IP AS asset_ip
-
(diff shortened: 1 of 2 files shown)


VatsalJagani (approved):

cyences_app_for_splunk/default/props.conf
```diff
@@ -553,7 +553,7 @@ EVAL-dest_name = mvappend(DNS, NETBIOS)
 # For Cyences Assets datamodel
 FIELDALIAS-asset_id = HOST_ID AS asset_id
 FIELDALIAS-asset_ip = IP AS asset_ip
-
+LOOKUP-qualys_kb_lookup = qualys_kb_lookup QID OUTPUT TITLE as description
```
mahirchavda, line 556: @VatsalJagani @hardikhdholariya The issue here is that if customer is not using qualys then it will generate lookup error. I faced same issue earlier for qualys iteself.

Can you make sure description field is not extracted by other. that might cause issue for CIM.

cyences_app_for_splunk/default/savedsearches.conf
```diff
@@ -2895,22 +2895,30 @@ alert.suppress = 0
...
 display.page.search.tab = statistics
 display.page.search.mode = fast
 request.ui_dispatch_app = cyences_app_for_splunk
 request.ui_dispatch_view = search
-search = `cs_sophos` type="Event::Endpoint::SavDisabled" \
-| stats count, latest(_time) as _time, values(src_ip) as src_ip, values(suser) as user by host, dhost \
-| eval cyences_severity = "high" \
+# Search triggers critical alerts if RealTime Protection is disabled for more than three hours, otherwise it triggers high severity alerts.
+# In order to avoid duplicate alerts having high severity, the search is running in 2 diff timeranges and merging its result.
+search = `cs_sophos` type="Event::Endpoint::Sav*" earliest=-4h@m latest=-2m@m \
```
mahirchavda, line 2913: @hardikhdholariya I'm not sure about the logic. wouldn't this generate duplicate resuls for all events where the sophos was disabled before last 3 hours?

cyences_app_for_splunk/default/props.conf
hardikhdholariya: Deleted.

mahirchavda (approved):

PR #362, opened 2023-08-14
Update the device inventory test cases

hardikhdholariya:
- Updated the test cases as per the device inventory implementation.

### DIFF
--- a/demo/test_device_inventory_v2.py
+++ b/demo/test_device_inventory_v2.py
@@ -17,6 +17,7 @@ from splunk import rest
 # Following things needs to be configured as per the splunk environment
 SPLUNK_USERNAME = "admin"
 SPLUNK_PASSWORD = "admin123"
+
 HOSTNAME_POSTFIXES = ".ad.crossrealms.com, .crossrealms.com"
 APP_NAME = "cyences_app_for_splunk"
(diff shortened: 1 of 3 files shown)


demo/test_device_inventory_v2.py
```diff
@@ -2,13 +2,24 @@
...
 from device_inventory_v2_util import DeviceEntry, DeviceManager
+from cyences_device_manager import logger
+import splunk.auth as auth
+from splunk import rest
 
 
+# Following things needs to be configured as per the splunk environment
+SPLUNK_USERNAME = "admin"
+SPLUNK_PASSWORD = "admin123"
 HOSTNAME_POSTFIXES = ".ad.crossrealms.com, .crossrealms.com"
```
VatsalJagani, line 17: Add new line above this as keep the SPLUNK_USERNAME and SPLUNK_PASSWORD those 2 are only user needs to configure, rest of the variables are static.

demo/test_device_inventory_v2.py
VatsalJagani, line 1: Add a comment on top of the file on how to execute these test cases as this needs to be executed with Splunk python cmd only.

VatsalJagani (approved):

PR #365, opened 2023-08-16
Workflow updated to use the latest action.

VatsalJagani:


.github/workflows/on_pr.yml
mahirchavda, line 1: I dont think we need on_pr. we can add condition in the on_master.yml file itself.

VatsalJagani: I thought this is more clear distinction between what gets executed and when.

mahirchavda (approved):

PR #374, opened 2023-08-23
CY-462: Filter the password expired events from the Bruteforce alert

mahirchavda:


cyences_app_for_splunk/default/macros.conf
```diff
@@ -1158,12 +1158,12 @@ definition = Authentication.app=*
 iseval = 0
 
 [cs_bruteforce_from_user_additional_filter]
-definition = ()
+definition = (Authentication.error_code!="0xC0000071")
```
VatsalJagani, line 1161: Can we please add a comment here on what this error code is about?

VatsalJagani (approved):

PR #376, opened 2023-08-23
Linux data collection doc improved.

VatsalJagani:


docs/data_onboarding/linux.md
```diff
@@ -19,69 +19,70 @@ Installation and Configuration Guide:
...
+    # Lists users who could login (i.e., they are assigned a login shell)
+    [script://./bin/usersWithLoginPrivs.sh]
+    sourcetype = usersWithLoginPrivs
+    source = usersWithLoginPrivs
+    interval = 3600
+    disabled = 0
     index = os
 
     # Shows stats per link-level Etherner interface (simply, NIC) 
     [script://./bin/interfaces.sh]
```
mahirchavda, line 38: I think most of these inputs requires root user permission

mahirchavda (approved):

docs/data_onboarding/linux.md
VatsalJagani: We have that note for TA-cyences section on the same page, so fine for now.

PR #382, opened 2023-08-28
Add host to the device inventory v2 from  internal logs

hardikhdholariya:
- Added the table border of the expanded tables
- Added savedsearch to add host to the device inventory v2 from internal logs


cyences_app_for_splunk/default/savedsearches.conf
```diff
@@ -3506,7 +3506,7 @@ search = `cs_lansweeper` `cs_lansweeper_timerange` \
 | eval product_name="Lansweeper", product_uuid=AssetID, hostname=lower(mvdedup(mvappend(AssetName, FQDN))), ip=lower(IPAddress), mac_address=lower(Mac), antivirus=mvzip(antivirus_name, antivirus_enabled, "#") \
 | rename _time as time, AssetID as lansweeper_id, host as lansweeper_collected_by, site_name as Site, AssetTypename as AssetType, Statename as lansweeper_state, Userdomain as Domain, AssetGroup as GroupName, OScode as OSVersion, Username as lansweeper_user, version as AssetVersion, OS as lansweeper_os, FQDN as lansweeper_fqdn, Firstseen as FirstSeen, Lastseen as LastSeen \
 | fillnull value="" hostname mac_address ip \
-| dedup hostname mac_address ip \
+| dedup hostname mac_address ip ```In order to avoid the duplicate entries of the similar devices which has diff AssetID``` \
```
VatsalJagani, line 3509: We need to remove this as we discussed.

cyences_app_for_splunk/default/savedsearches.conf
```diff
@@ -3705,6 +3705,27 @@ search = `cs_kaspersky` ProductName=KES*  \
...
+display.general.type = statistics
+display.page.search.tab = statistics
+display.page.search.mode = fast
+request.ui_dispatch_app = cyences_app_for_splunk
+request.ui_dispatch_view = search
+search = (index=_internal sourcetype=splunkd (connectionType=cooked OR connectionType=cookedSSL) fwdType=* group=tcpin_connections) \
+| stats latest(sourceIp) as ip, latest(fwdType) as forwarder_type, latest(version) as version, latest(arch) as arch, latest(os) as os, max(_time) as time values(splunk_server) as splunk_server by hostname \
+| eval product_name="Splunk Internal", product_uuid=hostname."-".ip, mac_address="" \
+| cyencesdevicemanager operation="addentries"
+action.cyences_notable_event_action.products = Splunk Internal
```
VatsalJagani, line 3727: @mahirchavda - Please review this and advice.

cyences_app_for_splunk/default/savedsearches.conf
```diff
@@ -3705,6 +3705,27 @@ search = `cs_kaspersky` ProductName=KES*  \
 | outputlookup cs_kaspersky_inventory_v2
 action.cyences_notable_event_action.products = Kaspersky
 
+[Device Inventory - Splunk Internal - V2]
```
VatsalJagani, line 3708: This should be similar to Lansweeper. Get the latest things only, and we can  delete old stuff.

cyences_app_for_splunk/default/savedsearches.conf
hardikhdholariya: Yes, we've discussed to remove it. but I found it useful in some scenarios as following:
If 2 assets with diff assetid and same host and empty mac & ip will be there then 
1) dedup will treat it as a one asset
2) without using dedup, it will be 2 assets in device inventory because it considers the same asset only if 2 property(from host, ip, mac) is same (where values should not be empty or null).

VatsalJagani: ok

VatsalJagani (approved):

cyences_app_for_splunk/appserver/static/js/build/cs_configuration.js
VatsalJagani, line 1: @mahirchavda  - I think this needs to be changed outside minified JS. Please work with @mahirchavda do make that change.

PR #388, opened 2023-09-13
Cy 500 implement device inventoy v2

hardikhdholariya:
- merging the device inventory v2 branch to master.


.gitignore
```diff
@@ -5,6 +5,8 @@ TA-cyences/local/**
 demo/dist/**
 demo/RansomwareDemo/**
 demo/RansomwareDemo.zip
+demo/device_list.pickle
```
VatsalJagani, line 8: please remove it, no longer in use.

cyences_app_for_splunk/default/data/ui/views/cs_user_inventory_table.xml
```diff
@@ -0,0 +1,153 @@
+<form theme="dark" version="1.1">
```
VatsalJagani, line 1: Doesn't this need Javascript like a device inventory dashboard?

demo/test_device_inventory_v2.py
VatsalJagani, line 1: Did you executed the tests manually at the end for final verification?

product_setup_page/src/allMacros.js
```diff
@@ -1,4 +1,12 @@
 let allMacros = [
+    {
+        section: 'Device Inventory V2',
```
VatsalJagani, line 3: Keep the section name as "Device and User Inventory V2"

VatsalJagani (approved):

demo/test_device_inventory_v2.py
hardikhdholariya: Updated the test cases as we added indextime and users field later on.
Also, clenaup logic updated based on indextime instead of time hence made changes in indextime for cleanup related test cases.

PR #390, opened 2023-09-14
corrections after testing

hardikhdholariya:
- removed commented code
- updated the saved search and macro
- minor UI changes in dashboard


cyences_app_for_splunk/default/macros.conf
```diff
@@ -710,7 +709,7 @@ definition = ".ad.crossrealms.com, .crossrealms.com, .local"
 
 # User Inventory
 [cs_user_inventory_user_postfixes]
-definition = ".ad.crossrealms.com, @crossrealms.com, .crossrealms.com, .local"
+definition = ".ad.crossrealms.com, @crossrealms.com, @CROSSREALMS.LOCAL, .crossrealms.com, .local"
```
VatsalJagani, line 712: Is this case insensitive, right? Just confirming.

VatsalJagani (approved):

cyences_app_for_splunk/default/macros.conf
hardikhdholariya: Yes, everything matching after converting it to lower case.

PR #395, opened 2023-09-18
Add exclusions in user inventory search

hardikhdholariya:
Following users are excluded from user inventory gen:
- users having value "' or '1=1",  "-" and empty
- sourcetype="WinRegistry" => process users ends with ".exe"
- sourcetype="WinEventLog" Account_Domain="NT VIRTUAL MACHINE" => exclude user having values like "5F2FBSE9-F8E7-42FF-974F-D51EF8F1DF8D"
- sourcetype="WinEventLog" user_type="computer" => exclude users ended with $
- In user inventory python script, excluded the users ending with $
- optimized the drilldown search


cyences_app_for_splunk/bin/user_inventory_v2_util.py
```diff
@@ -378,6 +392,9 @@ def add_user_entry(self, new_entry: UserEntry):
         # Assign a unique user_uuid to the user
         matching_user = self._find_user(new_entry)
 
+        if matching_user is None:
```
VatsalJagani, line 395: Logically this looks not right.

Use the functionis_user_ends_with separately and directly here.

cyences_app_for_splunk/bin/user_inventory_v2_util.py
```diff
@@ -229,6 +241,8 @@ def get_user_details(self):
         return _users
 
     def _find_user(self, user_entry: UserEntry):
+        if is_user_ends_with(user_entry.user, user_ends_with=["$"], user_postfixes=self.user_postfixes):
```
VatsalJagani, line 244: This function should only return boolean logically.

cyences_app_for_splunk/default/macros.conf
```diff
@@ -712,9 +712,9 @@ definition = ".ad.crossrealms.com, .crossrealms.com, .local"
 definition = ".ad.crossrealms.com, @crossrealms.com, @CROSSREALMS.LOCAL, .crossrealms.com, .local"
 # Put it with double quote on both side (compulsory)
 
-[cs_user_inventory_exclude_users]
-definition = NOT user IN ("$*")
-# To exclude the specific users, add it inside the brackets by comma separated
+[cs_user_inventory_exclude_data]
+definition = NOT (sourcetype="WinEventLog" Account_Domain="NT VIRTUAL MACHINE")
```
VatsalJagani, line 716: We should apply this to environment and not commit.

For the repo, use the value * maybe.

cyences_app_for_splunk/default/macros.conf
```diff
@@ -712,9 +712,9 @@ definition = ".ad.crossrealms.com, .crossrealms.com, .local"
...
+# To exclude the specific type of data, above example is to remove the users having values "5F2FBSE9-F8E7-42FF-974F-D51EF8F1DF8D"
```
VatsalJagani, line 717: And this comment is also related and needs to be removed.

VatsalJagani (changes requested):

cyences_app_for_splunk/bin/user_inventory_v2_util.py
hardikhdholariya: Yes, it returns only boolean.

cyences_app_for_splunk/bin/user_inventory_v2_util.py
```diff
@@ -376,6 +388,9 @@ def _create_new_user(self):
 
     def add_user_entry(self, new_entry: UserEntry):
         # Assign a unique user_uuid to the user
+        if is_user_ends_with(new_entry.user, user_ends_with=["$"], user_postfixes=self.user_postfixes):
+            return "n/a"
```
VatsalJagani, line 392: Why are we returning "n/a" instead of None or false or raise Exception?

product_setup_page/src/allMacros.js
```diff
@@ -4,7 +4,7 @@ let allMacros = [
         macros: [
             {name: 'cs_device_inventory_hostname_postfixes' },
             {name: 'cs_user_inventory_user_postfixes' },
-            {name: 'cs_user_inventory_exclude_users', description: 'To exclude the specific users, add it inside the brackets by comma separated. (e.g. NOT user IN ("user1", "user2"))' },
+            {name: 'cs_user_inventory_exclude_data', description: 'To exclude the users based on some fields. (e.g. NOT (field1=value1 field2=value2))' },
```
VatsalJagani, line 7: Let's have it as a common syntax as a filter.

search *

Update description according to that.

VatsalJagani (changes requested):

cyences_app_for_splunk/bin/user_inventory_v2_util.py
hardikhdholariya: return value will be displayed as user_id field (random uuid) on search output.
In case of user ends with $, we ignoring the entry hence it will display user_id as "n/a". In case of None and False, it will display as blank and 0 respectively.

VatsalJagani (approved):

PR #396, opened 2023-09-20
Exclude the users from user inventory

hardikhdholariya:
- Converted user field to multivalued and updated the saved search accordingly.
- Excluded users having unnecessary values like mac. from cisco meraki.
- Added the user inventory prefix macro.


cyences_app_for_splunk/default/savedsearches.conf
```diff
@@ -4086,12 +4086,14 @@ display.page.search.tab = statistics
...
 | `cs_user_inventory_data_filter` \
-| eval user=coalesce(src_user, user, user_name, username,Username, User, UserName, userName) \
-| search NOT user IN ("$*", "-", "' or '1=1", "") \
-| eval user_type=coalesce(user_type, user_role, user_category, "n/a") \
+| eval user=mvdedup(mvappend(coalesce(src_user, user, user_name, username,Username, User, UserName, userName), mail)), user_type=coalesce(user_type, user_role, user_category, "n/a") \
+| search NOT user IN ("$*", "-", "' or '1=1", ".{jndi", "") \
+| nomv user \
 | stats latest(user_type) as user_type values(index) as index values(sourcetype) as sourcetype latest(_indextime) as indextime by user vendor_product \
+| makemv delim="\
+" user \
```
VatsalJagani, line 4096: delim in multiline is it expected?

VatsalJagani (approved):

cyences_app_for_splunk/default/savedsearches.conf
hardikhdholariya: Yes, **nomv** command add the newline between each values hence while converting it to multivalue using **makemv** delim should be new line. 
Here delim is supporting string not the regex hence i added new line as a string.

PR #404, opened 2023-09-29
Added custom command to backfill the lookups

hardikhdholariya:
- Created custom command to run the enabled searches to backfill the device/user inventory.
- Updated user inventory backfill time range from 7 days to last 24 hours as it is taking much time.

### DIFF
--- a/cyences_app_for_splunk/bin/cyences_inventory_backfill.py
+++ b/cyences_app_for_splunk/bin/cyences_inventory_backfill.py
@@ -1,0 +1,65 @@
+#!/usr/bin/env python
+
+import sys
+import time
+
+from splunklib.searchcommands import (
+    dispatch,
+    GeneratingCommand,
+    Configuration,
+    Option,
+    validators,
+)
+
+import splunklib.client as client
+import splunklib.results as results
+import cs_utils
+import logging
+import logger_manager
+
+logger = logger_manager.setup_logging("inventory_backfill", logging.INFO)
+
+
+def handle_results(response):
+    reader = results.JSONResultsReader(response)
+    for result in reader:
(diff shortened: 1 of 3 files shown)


cyences_app_for_splunk/bin/cyences_inventory_backfill.py
```diff
@@ -0,0 +1,65 @@
...
+            self.conf_manger = cs_utils.ConfigHandler(logger, self.session_key)
+
+            saved_searches = self.conf_manger.get_saved_searches()
+
+            for saved_search in saved_searches:
+                name = saved_search["name"]
+                if name.startswith(self.search_prefix) and (not cs_utils.is_true(saved_search["content"].get("disabled"))):
+
+                    if name == "Device Inventory - Windows Defender":
+                        SEARCH_QUERY = '| search `cs_windows_defender` EventCode=1151 `cs_windows_defender_max_timerange` | `cs_windows_defender_inventory_fill_search` | where SEARCHNOTHING="SEARCHNOTHING"'
```
VatsalJagani, line 48: Why we have this query hard-coded?

VatsalJagani (approved):

cyences_app_for_splunk/bin/cyences_inventory_backfill.py
hardikhdholariya: it's to limit the time range to last 24 hour using this macro:: cs_windows_defender_max_timerange

don't know the actual reason but previously, it is defined like this and rest of the search is also in macro (cs_windows_defender_inventory_fill_search) so that it's easy to use at both places (in actual saved search and in backfill search)

mahirchavda (approved):

PR #406, opened 2023-10-03
Updated the app version

hardikhdholariya:
Updated the app version to 4.3.0

### DIFF
--- a/cyences_app_for_splunk/default/app.conf
+++ b/cyences_app_for_splunk/default/app.conf
@@ -12,7 +12,7 @@ label = Cyences App for Splunk
 [launcher]
 author = CrossRealms International
 description = Cyences App for Splunk built by CrossRealms International
-version = 4.2.1
+version = 4.3.0
 
 [package]
 id = cyences_app_for_splunk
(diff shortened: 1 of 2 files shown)


mahirchavda (approved):

cyences_app_for_splunk/default/app.conf
```diff
@@ -12,7 +12,7 @@ label = Cyences App for Splunk
 [launcher]
 author = CrossRealms International
 description = Cyences App for Splunk built by CrossRealms International
-version = 4.2.1
+version = 4.3.0
```
VatsalJagani, line 15: Just grep to see that's the place we have the version number mentioned, just to ensure.

VatsalJagani, line 15: Change version number in JS project for configuration page as well.

hardikhdholariya: Didn't get it? do you mean on cyences settings page anywhere? greped it but do not found.

VatsalJagani: https://github.com/CrossRealms/Splunk-Cyences-App-for-Splunk/blob/master/product_setup_page/package.json

hardikhdholariya: Discussed with mahir, it is not required as we are not using it anywhere.

VatsalJagani (approved):

PR #414, opened 2023-10-11
Added panels relaetd to the google workspace data

hardikhdholariya:
- Added panels related to the google workspace data
- Fixed o365 token issue

### DIFF
--- a/cyences_app_for_splunk/default/data/ui/views/cs_gsuite_reports.xml
+++ b/cyences_app_for_splunk/default/data/ui/views/cs_gsuite_reports.xml
@@ -254,7 +254,7 @@
         <fieldForLabel>event.name</fieldForLabel>
         <fieldForValue>event.name</fieldForValue>
         <search>
-          <query>`cs_gsuite` sourcetype="gws:reports:groups_enterprise" NOT "event.name" IN ("*member*", "*user*", "*join*") user=$tkn_user|s$ | dedup event.name | table event.name</query>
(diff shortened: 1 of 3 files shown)


cyences_app_for_splunk/default/data/ui/views/cs_gsuite_reports.xml
```diff
@@ -53,14 +57,34 @@
...
+    </panel>
+  </row>
   <row>
     <panel>
       <title>Login Challenged</title>
       <table>
         <title>User was asked for extra login question due to suspicious login activity</title>
         <search>
-          <query>`cs_gsuite_login_events` user=$tkn_user|s$ event_name="login_challenge" | sort login_challenge_status | table _time, user, ipAddress, login_type, login_challenge_method, login_challenge_status</query>
+          <query>`cs_gsuite_login_events` user=$tkn_user|s$ "event.name"="login_challenge" | sort login_type | `cs_human_readable_time_format(_time, Time)` | rename user as User, ipAddress as IPAddress, action as Action, login_challenge_method as LoginChallengeMethod, login_type as LoginType | table Time, User, IPAddress, Action, LoginChallengeMethod, LoginType</query>
```
VatsalJagani, line 87: login_challenge of google workspace is kind of similar to 2FA or MFA or Azure, so we create alerts and dashboard-panels according to that.

VatsalJagani (approved):

cyences_app_for_splunk/default/data/ui/views/cs_gsuite_reports.xml
hardikhdholariya: Yes, will check. it is kind of security question when you do the suspicious activity then google asks for the verification.

one of the method is by adding the Employee Id as a security question. Rahaf has tried to login from new devices but the question is not coming up. (https://support.google.com/a/answer/6002699?hl=en)

PR #440, opened 2023-11-02
Added colors to the panels

hardikhdholariya:


cyences_app_for_splunk/default/data/ui/views/cs_gsuite_reports.xml
VatsalJagani, line 1: Didn't we removed this dashboard?

VatsalJagani (changes requested):

cyences_app_for_splunk/default/data/ui/views/cs_gsuite_reports.xml
hardikhdholariya: file name is not changed. content is updated. becuase upgrade will not clean up the dashboard. should we change the filename?

VatsalJagani: I would say yes. Cleanup would happen for default directory I think.

VatsalJagani (approved):

PR #442, opened 2023-11-08
Added is_internal_top5_src_ip field to identify the internal/external ips

hardikhdholariya:


cyences_app_for_splunk/default/macros.conf
```diff
@@ -81,6 +81,11 @@ args = ip_field
 definition = ($ip_field$!="10.0.0.0/8" AND $ip_field$!="172.16.0.0/12" AND $ip_field$!="192.168.0.0/16")
 iseval = 0
 
+[cs_is_source_internal(1)]
```
VatsalJagani, line 84: Can't we use already exting macro cs_private_ips?

hardikhdholariya: IP masking search is not working inside the **eval** hence used the **cidrmatch**

VatsalJagani: Just keep the naming convetion consistent with previous macro.

VatsalJagani (approved):

PR #444, opened 2023-11-09
Add privilege option for user in user inventory

hardikhdholariya:
- Added privilege option for user in user inventory
- Fixed the click event issue on forensic dashboard. (action should be taken when click on the particular element (ex. checkbox) of the cell. it should not allow if user clicks anywhere in cell outside the checkbox.)

### DIFF
--- a/cyences_app_for_splunk/appserver/static/notable_event_editor.js
+++ b/cyences_app_for_splunk/appserver/static/notable_event_editor.js
@@ -406,10 +406,10 @@ require([
                     } else {
                         $td.html('<input type="checkbox" class="notable_event_selector" id="notable_event_selector" name="notable_event_selector" value="' + cell.value + '"></input>');
                     }
-                    $td.on("click", function (e) {
-                        e.stopPropagation();
-                        $td.trigger("cyences_notable_event_handlers", { "action": cell.field });
-                    });
+                    $td.children("input").on("click", function (e) {
+                      e.stopPropagation();
+                      $td.trigger("cyences_notable_event_handlers", { "action": cell.field });
+                  });
                 }
                 else {
                     $td.addClass(cell.field).html("");
@@ -435,7 +435,7 @@ require([
 
                     $td.children('[data-toggle="tooltip"]').tooltip();
 
-                    $td.on("click", function (e) {
+                    $td.children("a").on("click", function (e) {
                         console.log("event handler fired");
                         e.stopPropagation();
                         $td.trigger("cyences_notable_event_handlers", { "action": cell.field });
--- a/cyences_app_for_splunk/appserver/static/privilege_user_role_editor.js
+++ b/cyences_app_for_splunk/appserver/static/privilege_user_role_editor.js
@@ -1,0 +1,289 @@
+require([
+    "splunkjs/mvc",
+    "underscore",
+    "jquery",
+    'splunkjs/mvc/tableview',
+    '../app/cyences_app_for_splunk/splunk_common_js_v_utilities'
+], function (
+    mvc,
+    _,
+    $,
+    TableView,
+    SplunkCommonUtilities
+) {
+
+    const USER_INVENTORY_TABLE_SEARCH_ID = 'base_summary';
+    const USER_INVENTORY_TABLE_ID = 'user_inventory_main_tbl';   // NOTE - CSS contains this hard-coded value.
+    const DEFAULT_USER_ROLE = '(unchanged)';
+    
+    let AVAILABLE_USER_ROLES = [DEFAULT_USER_ROLE, 'Administrator', 'admin', 'Dcadmin', 'Root'];
+    let selected_user_row_events = [];
+    let all_users = [];
+
+    // creates the div having diff options to select/edit the raws if the search contains the results.
+    let userEventSearchManager = new SplunkCommonUtilities.VSearchManagerUtility(
+        function(data){
+            if (data == null){
+                all_users = [];
+                $("#bulk_edit_container").remove();
+                return;
+            }
+            let uuid = data.fields.indexOf("uuid");
+            all_users = _.map(data.rows, function (e) { return e[uuid]; });
+            $("#bulk_edit_container").remove();
+            $(`#${USER_INVENTORY_TABLE_ID}`).parent().before($("<div />").attr('id', 'bulk_edit_container').addClass("bulk_edit_container").addClass('panel-element-row'));
+            var links = _.template('<a href="#" id="bulk_edit_select_all">Select All</a> | <a href="#" id="bulk_edit_selected">Edit Selected</a> | <a href="#" id="bulk_edit_all">Edit All <%-nr_users%> Matching Users</a> | <a href="#" id="bulk_edit_clear">Reset Selection</a>', { nr_users: all_users.length });
+            $("#bulk_edit_container").html(links);
+            $("#bulk_edit_container").show();
+        },
+        function(errorProperties){
+            console.log("User inventory table search failed.");
+        }
+    );
+    userEventSearchManager.searchById(USER_INVENTORY_TABLE_SEARCH_ID);
+
+    function restartUserEventSearch(){
+        userEventSearchManager.startSearch();
+    }
+
+    // create the search to execute when user click on save button of the popup.
+    function runUserInventoryUpdaterSearch(entries_to_update, callBackFunction){
+        // Splitting array into multiple queries in case bulk update contains many results, it may result into failure
+        const chunkSize = 500;
+        let searchesCompleted = {};
+        for (let i = 0; i < entries_to_update.length; i += chunkSize) {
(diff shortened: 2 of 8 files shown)


cyences_app_for_splunk/appserver/static/notable_event_editor.js
```diff
@@ -406,10 +406,10 @@ require([
                     } else {
                         $td.html('<input type="checkbox" class="notable_event_selector" id="notable_event_selector" name="notable_event_selector" value="' + cell.value + '"></input>');
                     }
-                    $td.on("click", function (e) {
-                        e.stopPropagation();
-                        $td.trigger("cyences_notable_event_handlers", { "action": cell.field });
-                    });
+                    $td.children("input").on("click", function (e) {
```
VatsalJagani, line 409: This will limit where use can click only on checkbox not on full cell.

cyences_app_for_splunk/appserver/static/notable_event_editor.js
```diff
@@ -435,7 +435,7 @@ require([
 
                     $td.children('[data-toggle="tooltip"]').tooltip();
 
-                    $td.on("click", function (e) {
+                    $td.children("a").on("click", function (e) {
```
VatsalJagani, line 438: This will limit where use can click only on checkbox not on full cell.

cyences_app_for_splunk/appserver/static/privilege_user_role_editor.js
```diff
@@ -0,0 +1,289 @@
...
+        let searchesCompleted = {};
+        for (let i = 0; i < entries_to_update.length; i += chunkSize) {
+            const chunk = entries_to_update.slice(i, i + chunkSize);
+
+            let user_role = chunk[0].user_role;
+            let searchQuery = '';
+            for(let i=0; i<chunk.length; i++){
+                let entry = chunk[i];
+                if (i==0){
+                  searchQuery = `| inputlookup cs_user_inventory | search uuid IN (`;
```
VatsalJagani, line 62: Can we make use of custom command? So we can avoid inputlookup and outputlookup with device inventory and user inventory?

cyences_app_for_splunk/appserver/static/privilege_user_role_editor.js
```diff
@@ -0,0 +1,289 @@
...
+    // create the popup dialoge box when user wants to edit the entries.
+    function handlerUserInventoryEdit(handlerObj, data){
+        console.log("user_event_edit handler.");
+
+        let modal_title = "Privilege User Role";
+        let nr_user_events = data.user_event_uuids.length;
+        let user_event_uuids_string = data.user_event_uuids.join(':');
+
+        var edit_panel = '' +
+            '<div class="modal fade modal-wide shared-alertcontrols-dialogs-editdialog in" id="edit_panel">' +
```
VatsalJagani, line 118: As this dialog box html is pretty much constant please add the whole modal as part of a HTML separately only once when the JS loads.

And then from this function just show/hide the modal after changing necessary values.

cyences_app_for_splunk/appserver/static/privilege_user_role_editor.js
```diff
@@ -0,0 +1,289 @@
...
+        var user_event_uuids = $("#user_event_uuids").val().split(':');
+        var user_role = $("#user_role").val();
+        
+        let entries_to_update = [];
+        for(let i=0; i<user_event_uuids.length; i++){
+            let entry = {'uuid': user_event_uuids[i], 'user_role': user_role};
+            entries_to_update.push(entry);
+        }
+
+        runUserInventoryUpdaterSearch(entries_to_update, function(){
```
VatsalJagani, line 196: Function name seems to be mismatched compare to the code of the function.

cyences_app_for_splunk/appserver/static/privilege_user_role_editor.js
```diff
@@ -0,0 +1,289 @@
...
+                }
+                $td.children("input").on("click", function (e) {
+                    e.stopPropagation();
+                    $td.trigger("cyences_user_event_handlers", { "action": cell.field });
+                });
+            }
+            }
+        });
+
+    let CyencesUserEventCSSClassRenderer = TableView.BaseCellRenderer.extend({
```
VatsalJagani, line 265: Why do we need two separate cell renderers?

cyences_app_for_splunk/bin/user_inventory_util.py
```diff
@@ -210,7 +210,8 @@ def get_as_dict(self, user):
             'user_types': list(user.get("user_types").keys()),
             'indexes': list(user.get("indexes").keys()),
             'sourcetypes': list(user.get("sourcetypes").keys()),
-            'users': list(user.get("users").keys())
+            'users': list(user.get("users").keys()),
+            'privilege_user_role': str(user.get("privilege_user_role", "-"))
```
VatsalJagani, line 214: call it **is_previleged_user**, to make it more like a boolean field in the User inventory as well in the dashboard.

cyences_app_for_splunk/default/data/ui/views/cs_user_inventory_table.xml
```diff
@@ -129,18 +129,18 @@
   <row>
     <panel>
       <title>User Inventory Table</title>
-      <table>
+      <table id="user_inventory_main_tbl">
         <search base="base_summary">
           <query> | search $tkn_sp_filter$
 | rename users as User, user_types as "User Types", product_names as Products
+| table user_row_selector, User, "User Types",Products, uuid
```
VatsalJagani, line 136: add the field hear as as well is_previledged_user
* but combine with manually assigned value with one extracted from UserTypes.

VatsalJagani (changes requested):

cyences_app_for_splunk/appserver/static/notable_event_editor.js
hardikhdholariya: yes, it will limit only checkbox to clickable.

cyences_app_for_splunk/appserver/static/privilege_user_role_editor.js
hardikhdholariya: To update the lookup, we have to provide the same values for all the existing fields and values are in nested json format due to matching users. 
but custom command fetches the data by simplifying the json structure. 
and if we do with custom command then it is hard to decide which value pickup from multivalue field to again add it via custom command.

cyences_app_for_splunk/appserver/static/privilege_user_role_editor.js
hardikhdholariya: make sense. I'll combine it.

cyences_app_for_splunk/appserver/static/privilege_user_role_editor.js
VatsalJagani: We can pass uuid to the custom command and have a specific handler to handle the update is_previledged_user field right?

Similarly how we are making updates in Device Inventory. I'm trying to keep things consitent.

cyences_app_for_splunk/bin/user_inventory_util.py
hardikhdholariya: FYI, this field will display the role name which user has provided. are you expecting the yes or no on UI?

cyences_app_for_splunk/appserver/static/privilege_user_role_editor.js
hardikhdholariya: Yes we can do by adding the specific handler in user inventory. I'll add that.

cyences_app_for_splunk/default/macros.conf
```diff
@@ -23,7 +23,7 @@ definition = join notable_event_id type=left [| inputlookup cyences_notable_even
 | fillnull notable_event_status value="Unassigned"\
 | eval notable_event_assignee=if(isnotnull(notable_event_id) AND notable_event_id!="", notable_event_assignee, "BANANABANANA")\
 | eval notable_event_status=if(isnotnull(notable_event_id) AND notable_event_id!="", notable_event_status, "BANANABANANA")\
-| eval notable_event_selector=if(isnotnull(notable_event_id) AND notable_event_id!="", "-", "BANANABANANA")\
+| eval notable_event_selector=if(isnotnull(notable_event_id) AND notable_event_id!="", notable_event_id, "BANANABANANA")\
```
VatsalJagani, line 26: Was this a bug?

VatsalJagani (approved):

cyences_app_for_splunk/default/macros.conf
hardikhdholariya: Yes, if you select the checkbox on first page then go to second page and then come to first page then selected checkboxes will deselected automatically but on edit it will make change for that id. 
so it's UI rendering issue.

VatsalJagani (approved):

PR #466, opened 2023-12-08
Updated o365 AD alerts with azure data source

hardikhdholariya:


cyences_app_for_splunk/default/props.conf
```diff
@@ -94,6 +94,21 @@ EVAL-app = "AzureActiveDirectory"
 LOOKUP-cs_o365_error_codes = cs_o365_error_codes ErrorCode AS status.errorCode OUTPUT LogonError
 LOOKUP-cs_country_code = cs_country_code code AS location.countryOrRegion OUTPUT country AS org_country
 
+######################
+#### Azure AD ########
+######################
+[azure:aad:audit]
```
VatsalJagani, line 100: Nice job on field aliases. It simplify queries.

VatsalJagani (commented):
Changes looks good. But we need to think of two things:
* Important fields as old_ and new_ separate value.
* Update severity based on changes happened on these important fields changes.

VatsalJagani (approved):

PR #467, opened 2023-12-15
Add data reviewer dashboard

hardikhdholariya:
- Added custom commad to manage the data reviewer dashboard searches
- Added data reviewer dashboard
- Updated AWS and GWS product details

### DIFF
--- a/cyences_app_for_splunk/bin/cyences_product_manager.py
+++ b/cyences_app_for_splunk/bin/cyences_product_manager.py
@@ -1,0 +1,137 @@
+#!/usr/bin/env python
+
+import sys
+
+from splunklib.searchcommands import (
+    dispatch,
+    GeneratingCommand,
+    Configuration,
+    Option,
+)
+from cs_product_list import PRODUCTS
+
+import copy
+import re
+import logging
+import logger_manager
+
+logger = logger_manager.setup_logging("cyences_product_manager", logging.INFO)
+
+
+@Configuration()
+class CreateProductSpecificSearch(GeneratingCommand):
+    operation = Option(name="operation", require=True, default="getproducts")
+    product_name = Option(name="product_name", require=False, default="")
+
+    def validate_inputs(self):
+        if self.operation not in ["getproducts", "buildproductspecificsearches"]:
+            raise Exception(
+                "operation - allowed values: getproducts, buildproductspecificsearches"
+            )
+
+    def get_sources_and_values(self, text):
+        regex_pattern = r'\|\seval\s([^\=]*)\=split\("([^"]*)",\s","\)'
+        match = re.search(regex_pattern, text)
+
+        if match:
+            source_field = match.group(1)
+            values = match.group(2)
+
+        return source_field, values
+
+    def build_source_reviewer_search(self, source_field, values, invocation_no):
+        search = ""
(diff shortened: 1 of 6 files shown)


cyences_app_for_splunk/default/data/ui/nav/default.xml
```diff
@@ -41,6 +41,7 @@
     <view name="cs_decommision_host" />
     <view name="cs_sysmon_deploy_audit" /><!--Sysmon-->
     <view name="cs_o365_defender_atp_audit" /><!--Office 365 Defender ATP-->
+    <view name="cs_data_reviewer" />
```
VatsalJagani, line 44: Move it after configuration page.

cyences_app_for_splunk/default/data/ui/views/cs_data_reviewer.xml
```diff
@@ -0,0 +1,164 @@
...
+          <query>$tkn_metadata_count_search$ by sourcetype 
+| stats count</query>
+          <earliest>$timeRange.earliest$</earliest>
+          <latest>$timeRange.latest$</latest>
+        </search>
+        <option name="drilldown">none</option>
+        <option name="refresh.display">progressbar</option>
+      </single>
+    </panel>
+  </row>
```
VatsalJagani, line 88: Add new row with 3 tables:
* list of hosts
* list of source
* list of sourcetype

(max row count per page 5)

cyences_app_for_splunk/bin/cyences_product_manager.py
```diff
@@ -0,0 +1,137 @@
...
+import copy
+import re
+import logging
+import logger_manager
+
+logger = logger_manager.setup_logging("cyences_product_manager", logging.INFO)
+
+
+@Configuration()
+class CreateProductSpecificSearch(GeneratingCommand):
```
VatsalJagani, line 22: Keep the class name consistent with CyencesProductManager

cyences_app_for_splunk/bin/cyences_product_manager.py
```diff
@@ -0,0 +1,83 @@
...
+
+            # list down the product names to use in data reviewer dashboard filter
+            if self.operation == "getproducts":
+                for product in all_products:
+                    yield {"products": product["name"]}
+            # create product specific searches used in data reviewer dashboard
+            elif self.operation == "buildproductspecificsearches":
+                host_reviewer_search = ""
+                sources_reviewer_search = ""
+                metadata_count_search = "| tstats count where index=* "
```
VatsalJagani, line 45: Shall we also do metadata_count_search also in cs_product_list.py

VatsalJagani (changes requested):

cyences_app_for_splunk/bin/cyences_product_manager.py
hardikhdholariya: metadata_count_search = "| tstats count where index=* " ==> this is one time prefix.
the **repetitive part** is the source and sourcetype filter. we can add this to the cs_product_list but in that case we have to add product specific condition in custom command for the non tstats search (because it will not have the same prefix -> line no. 45).

VatsalJagani (approved):

PR #470, opened 2023-12-18
Add azure active directory data onboarding page

hardikhdholariya:


docs/data_onboarding/cloud_tenancies/o365_azure_ad.md
```diff
@@ -0,0 +1,24 @@
+---
+layout: default
+title: O365 Azure Active Directory
+permalink: /data_onboarding/cloud_tenancies/o365_azure_ad/
+nav_order: 5
```
VatsalJagani, line 5: Do we have to change the nav order for pages after this?

VatsalJagani (approved):

docs/data_onboarding/cloud_tenancies/o365_azure_ad.md
hardikhdholariya: It's the latest number but i think we have to add as per the order of file hence updating it,

VatsalJagani (approved):

PR #487, opened 2024-01-12
updated the app version

hardikhdholariya:


product_setup_page/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "setup_page",
-  "version": "1.0.3",
+  "version": "1.0.4",
```
VatsalJagani, line 3: I think we need to do 1.1.0 as this is not just bug fixes.

VatsalJagani (changes requested):

VatsalJagani (approved):

PR #493, opened 2024-01-22
Add subject prefix input to add on alert's email subject

hardikhdholariya:


product_setup_page/src/CyencesGeneralConfiguration.jsx
```diff
@@ -0,0 +1,87 @@
...
+import Heading from '@splunk/react-ui/Heading';
+import { generateToast } from './utils/util';
+import { axiosCallWrapper } from './utils/axiosCallWrapper';
+import ControlGroup from '@splunk/react-ui/ControlGroup';
+import Button from '@splunk/react-ui/Button';
+import Text from '@splunk/react-ui/Text';
+
+
+const EmailConfigurationFields = {
+    emailSubjectLabel: 'Email Subject Prefix',
```
VatsalJagani, line 11: Let's keep the name as **Environment Name**

product_setup_page/src/CyencesGeneralConfiguration.jsx
```diff
@@ -0,0 +1,87 @@
...
+    emaiSubjectHelp: 'Text to add as a prefix to email subject. For best practice, it is recommended to enclose text in brackets. E.g. [CrossRealms]',
```
VatsalJagani, line 12: Help -> This will be used in the alert emails sent by Cyences and other places.

cyences_app_for_splunk/default/alert_actions.conf
```diff
@@ -32,4 +33,5 @@ param.email_to_default =
 param.cyences_severities = critical
 param.email_to_include = 
 param.email_to_exclude = 
+param.subject_prefix = [N/A]
```
VatsalJagani, line 36: Because its a generic configuration, we need to put into some other file I think.

We can use not only for email but other places, so lets not store here.

cyences_app_for_splunk/README/alert_actions.conf.spec
```diff
@@ -6,10 +6,12 @@ param.filter_macro_value = <str> Filter macro value
 param.email_to = <str> Comma separated email addresses
 param.cyences_severities =  <str> Comma separated severities. ie. critical,high,medium,low,info
 param.exclude_alerts = <str> Comma separated alert names to exclude from digest email
+param.subject_prefix = <str> Text to add as a prefix of email subject
```
VatsalJagani, line 9: Because its a generic configuration, we need to put into some other file I think.

We can use not only for email but other places, so lets not store here.

cyences_app_for_splunk/bin/cyences_send_email.py
```diff
@@ -36,12 +38,15 @@ def transform(self, records):
...
 
+            logger.info("Time taken to fetch the final configurations = {} seconds".format(time.time() - start_time))
+
             logger.debug("Final alert action config: {}".format(alert_action_config))
 
             email_to_default = cs_utils.convert_to_set(alert_action_config.get("param.email_to_default"))
             cyences_severities = cs_utils.convert_to_set(alert_action_config.get("param.cyences_severities"))
             email_to_exclude = cs_utils.convert_to_set(alert_action_config.get("param.email_to_exclude"))
             email_to_include = cs_utils.convert_to_set(alert_action_config.get("param.email_to_include"))
+            subject_prefix = alert_action_config.get("param.subject_prefix", '') + " "
```
VatsalJagani, line 49: Add [] square brackets here instead of asking user to add it to config.

cyences_app_for_splunk/bin/cyences_send_digest_email.py
```diff
@@ -180,16 +183,16 @@ def transform(self, records):
 
                     if html_body.strip() != '':
                         if len(list_of_result_chunks) <= 1:
-                            subject = self.alert_name
+                            subject = subject_prefix + self.alert_name
                         else:
-                            subject = '{} Part-{}'.format(self.alert_name, email_counter)
+                            subject = '{}{} Part-{}'.format(subject_prefix, self.alert_name, email_counter)
```
VatsalJagani, line 188: Add [] square brackets here instead of asking user to add it to config.

VatsalJagani (changes requested):

VatsalJagani (changes requested):
Also full subject needs to be updated to: **Cyences Alert: [Environment] <Alert Name>** and **Cyences Alert Digest: [Environment] <Alert Name>**

Update the util file to remove from adding subject prefix.

cyences_app_for_splunk/bin/cyences_email_utility.py
```diff
@@ -269,9 +269,10 @@ def filter_recipients_if_not_in_allowed_domains(self, recipients):
             return recipients
 
 
-    def send(self, to, cc=[], bcc=[], subject='Splunk Alert', html_body=''):
+    def send(self, to, cc=[], bcc=[], subject='Cyences Alert', html_body='', is_digest=False):
```
VatsalJagani, line 272: Let's not make any change to subject under the email util file. Pass the full subject from calling file.

VatsalJagani (changes requested):

VatsalJagani (approved):

PR #541, opened 2024-04-10
Support msssql window eventlog format. Update user field

mahirchavda:
* support msswl window eventlog format
* update user fields


cyences_app_for_splunk/default/transforms.conf
```diff
@@ -203,6 +203,21 @@ SOURCE_KEY = ObjectGUID
 REGEX = (.+)
 FORMAT = objectGUID::$1
 
+# MSSQL
+[mssql_audit_event]
+SOURCE_KEY = Message
+REGEX = Audit event:\s(?<mssql_audit_event>.*)
+
+[mssql_audit_fields]
```
VatsalJagani, line 211: Do we need to extract the statement field explicitly?

VatsalJagani (approved):

cyences_app_for_splunk/default/transforms.conf
mahirchavda: No. It will be extracted as part of this

hardikhdholariya (approved):

PR #548, opened 2024-04-17
Add digest email configuration to setup page

hardikhdholariya:


mahirchavda (approved):

product_setup_page/src/SendEmailSetup.jsx
```diff
@@ -15,9 +14,18 @@ const EmailConfigurationFields = {
     passwordHelp: 'comma separated list of alert severity levels that are included in the email; default - critical severity only',
 }
 
+const DigestEmailConfigurationFields = {
+    usernameLabel: 'Default Email Recipients',
```
VatsalJagani, line 18: Is it necessary to keep username and password as field name?

PR #555, opened 2024-04-29
Excluding-NotAvailable-Users-o365Alerts

roaaattalla:
Filter the user "not available" events from the o365 login related alerts in the Cyences

### DIFF
--- a/cyences_app_for_splunk/default/savedsearches.conf
+++ b/cyences_app_for_splunk/default/savedsearches.conf
@@ -2326,11 +2326,11 @@ display.page.search.tab = statistics
 display.page.search.mode = fast
 request.ui_dispatch_app = cyences_app_for_splunk
 request.ui_dispatch_view = search
-search = `cs_o365` sourcetype="o365:management:activity" _index_earliest=-61m@m _index_latest=-1m@m Workload=AzureActiveDirectory Operation=UserLoginFailed (LogonError="DeviceAuthenticationRequired" OR LogonError="UserStrongAuthClientAuthNRequiredInterrupt") \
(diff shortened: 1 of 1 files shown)


VatsalJagani:
@roaaattalla , @hardikhdholariya  - Please ensure that we have to remove the logic for severity change as well as in certain cases we added to reduce the severity for "not available" user.

Also, in some alert even this user may be relavent. So ensure we are not filter if that's the case for any alert.

cyences_app_for_splunk/default/savedsearches.conf
```diff
@@ -2326,11 +2326,11 @@ display.page.search.tab = statistics
...
+search = `cs_o365` sourcetype="o365:management:activity" _index_earliest=-61m@m _index_latest=-1m@m Workload=AzureActiveDirectory Operation=UserLoginFailed (LogonError="DeviceAuthenticationRequired" OR LogonError="UserStrongAuthClientAuthNRequiredInterrupt") user!="not available"\
 | fields _time sourcetype user ClientIP Id LogonError user_type authentication_method app ApplicationId ExtendedProperties{}.Name ExtendedProperties{}.Value \
 | eval ExtendedProperties=mvzip('ExtendedProperties{}.Name','ExtendedProperties{}.Value'," : ") \
 | append \
-    [| search `cs_o365` sourcetype=o365:graph:api source="AuditLogs.SignIns" earliest=-2h@m latest=now status.errorCode!=0 (LogonError="DeviceAuthenticationRequired" OR LogonError="UserStrongAuthClientAuthNRequiredInterrupt") \
+    [| search `cs_o365` sourcetype=o365:graph:api source="AuditLogs.SignIns" earliest=-2h@m latest=now status.errorCode!=0 (LogonError="DeviceAuthenticationRequired" OR LogonError="UserStrongAuthClientAuthNRequiredInterrupt") user!="not available" \
```
hardikhdholariya, line 2333: Make sure that user field is available under these source types if not available then use **NOT user="not available"** instead of user!="not available" otherwise it remove all events.

PR #563, opened 2024-05-20
Convert time to local time

roaaattalla:


cyences_app_for_splunk/default/savedsearches.conf
```diff
@@ -580,6 +580,7 @@ search = `cs_windows_defender` EventCode IN(1006, 1007, 1008, 1015, 1116, 1117,
 | table Detection_Time, Computer, EventCodeDescription, Threat_Name, severity, cyences_severity, action, Category_Name, Detection_User, Additional_Actions_String, Threat_ID, Detection_ID, Process_Name, Path, is_privileged_user \
 | rename Additional_Actions_String as Additional_Action \
 | `cs_windows_defender_malware_detected_filter`
```
VatsalJagani, line 582: Missing backward slash at the end.

cyences_app_for_splunk/default/macros.conf
```diff
@@ -1232,6 +1232,9 @@ iseval = 0
 definition = search *
 iseval = 0
 
+[cs_windows_defender_malware_detected_local_time_converter]
```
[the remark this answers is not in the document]
roaaattalla: That detection time field is already human readable.
Since we have to convert it to local time zone we are first parsing it and then converting it to local time, that why we are using this new macro

cyences_app_for_splunk/default/macros.conf
```diff
@@ -1232,6 +1232,9 @@ iseval = 0
...
+definition = eval Detection_Time = strftime(strptime(Detection_Time, "%Y-%m-%dT%H:%M:%S.%QZ"), "%F %T %Z")
```
hardikhdholariya, line 1236: Make this macro generic by adding 2 params:: field name and format of the timestamp

definition = eval "$fieldname$" = strftime(strptime('$fieldname$', "$timeformat$"), "%F %T %Z")

Macro name should be something like:: cs_local_time_conversion

cyences_app_for_splunk/default/savedsearches.conf
```diff
@@ -579,7 +579,8 @@ search = `cs_windows_defender` EventCode IN(1006, 1007, 1008, 1015, 1116, 1117,
 | eval action = if(mvcount(action)>1, mvfilter(!match(action,"deferred")), action) \
 | table Detection_Time, Computer, EventCodeDescription, Threat_Name, severity, cyences_severity, action, Category_Name, Detection_User, Additional_Actions_String, Threat_ID, Detection_ID, Process_Name, Path, is_privileged_user \
 | rename Additional_Actions_String as Additional_Action \
-| `cs_windows_defender_malware_detected_filter`
+| `cs_windows_defender_malware_detected_filter` \
+| `cs_windows_defender_malware_detected_local_time_converter`
```
hardikhdholariya, line 583: keep this macro before the filter macro

cyences_app_for_splunk/default/macros.conf
hardikhdholariya, line 1235: Keep this macro under # Time section in macros.conf

hardikhdholariya (changes requested):

cyences_app_for_splunk/default/macros.conf
hardikhdholariya: `cs_local_time_conversion(Detection_Time, "%Y-%m-%dT%H:%M:%S.%QZ")`

VatsalJagani (approved):

hardikhdholariya (approved):

PR #571, opened 2024-06-05
Use sophos endpoint data for device inventory

hardikhdholariya:


cyences_app_for_splunk/bin/cs_product_list.py
```diff
@@ -131,7 +131,7 @@ def build_source_reviewer_search(by, values, first_call=True):
                 "macro_name": "cs_sophos",
                 "label": "Sophos Endpoint Protection Data",
                 "search_by": "sourcetype",
-                "search_values": "sophos_events",
```
VatsalJagani, line 134: Create a separate PR for documentation changes for this data collection and upgrade guide.

cyences_app_for_splunk/default/savedsearches.conf
```diff
@@ -6517,13 +6517,6 @@ search = index=* sourcetype!="WinRegistry" NOT (sourcetype="WinEventLog" user_ty
 | stats latest(user_type) as user_type values(index) as index values(sourcetype) as sourcetype latest(_indextime) as indextime by user vendor_product \
 | makemv delim="\
 " user \
-| append \
```
VatsalJagani, line 6520: Why we have removed this append??

hardikhdholariya: Previously, we were appending sophos users via custom command.

but now we are collecting data from sophos_endpoint source hence those logs will be covered in main search and for that I've written required extraction in props.conf

VatsalJagani (approved):

PR #578, opened 2024-06-11
alert configuration doc changes

hardikhdholariya:


VatsalJagani (changes requested):
Add a upgrade guide to check on all disabled alert that customer never wants to enable, and update this parameter.

VatsalJagani (approved):
Approving the PR but do not merge.

PR #584, opened 2024-06-20
Added splunk fwdtype and os details to the device inventory

hardikhdholariya:


cyences_app_for_splunk/appserver/static/cs_device_inventory_table.js
```diff
@@ -209,7 +213,7 @@ require([
...
             | rename ips as ip, hostnames as hostname, mac_addresses as mac_address
-            | table uuid, Select, _time, ip, hostname, mac_address, lansweeper_id, lansweeper_state, lansweeper_asset_type, lansweeper_os, lansweeper_user, lansweeper_description, qualys_id, QUALYS_OS, qualys_network_id, tenable_uuid, tenable_os, nessus_uuid, nessus_os, sophos_uuid, sophos_type, sophos_os, sophos_user, sophos_login_via, sophos_health, sophos_product_installed, crowdstrike_id,kaspersky_collected_by,kaspersky_version,kaspersky_host, kaspersky_status windows_defender_host 
+            | table uuid, Select, _time, ip, hostname, mac_address, lansweeper_id, lansweeper_state, lansweeper_asset_type, lansweeper_os, lansweeper_user, lansweeper_description, qualys_id, QUALYS_OS, qualys_network_id, tenable_uuid, tenable_os, nessus_uuid, nessus_os, sophos_uuid, sophos_type, sophos_os, sophos_user, sophos_login_via, sophos_health, sophos_product_installed, crowdstrike_id,kaspersky_collected_by,kaspersky_version,kaspersky_host, kaspersky_status, windows_defender_host, splunk_host, splunk_os, splunk_user, splunk_version, splunk_forwarder_type, splunk_server
```
VatsalJagani, line 216: Do we need splunk_server?

What's the value for splunk_user? Is it multi-valued field?

cyences_app_for_splunk/default/macros.conf
```diff
@@ -239,12 +243,15 @@ definition = cyencesdevicemanager operation="getdevices" \
 | eval kaspersky_status=kaspersky_status \
 | eval tmp_kaspersky_status = kaspersky_status, kaspersky=case(tmp_kaspersky_status == "Present", 0, tmp_kaspersky_status == "Disabled", 1) \
 | fieldformat kaspersky=tmp_kaspersky_status \
+| eval splunk_status=case(isnull(splunk_last_event), "-", 1==1, splunk_forwarder_type) \
```
VatsalJagani, line 246: splunk_status word is misleading as one would expect to have up/down status.

Instead we can say splunk_type maybe.

cyences_app_for_splunk/default/data/ui/views/cs_asset_intelligence.xml
VatsalJagani, line 1: Shall we add this as a separate table to show all the Splunk related fields?

VatsalJagani (changes requested):

cyences_app_for_splunk/appserver/static/cs_device_inventory_table.js
hardikhdholariya: splunk_server is not much useful as we have hostname hence removing it.

I forgot to remove splunk_user from here as no such info available. I'll remove it.

cyences_app_for_splunk/default/data/ui/views/cs_asset_intelligence.xml
hardikhdholariya: Sure, will add same as other product table to show/hide when it's related to splunk.

VatsalJagani (approved):

PR #587, opened 2024-06-21
added upgrade steps

hardikhdholariya:


cyences_app_for_splunk/bin/upgrade_steps.py
```diff
@@ -92,6 +92,14 @@ def upgrade_4_8_0(session_key, logger):
     time.sleep(60)
 
 
+def upgrade_4_9_0(session_key, logger):
+    service = client.connect(token=session_key, app=cs_utils.APP_NAME)
+
+    SPLUNK_DEVICES_CLEANUP_SEARCH = '| makeresults count=1 | eval time=now() | map search="| cyencesdevicemanager operation="cleanup" products_to_cleanup="Splunk Internal" minindextime=$time$"'
```
VatsalJagani, line 98: Why do we have to use map command?

hardikhdholariya: in order to use the latest time value inside custom command from **time** field.

VatsalJagani (approved):

PR #591, opened 2024-06-27
Port search filter added to Network Telemetry dashboard.

VatsalJagani:


mahirchavda (approved):

cyences_app_for_splunk/default/data/ui/views/cs_network_reports.xml
```diff
@@ -30,6 +30,10 @@
       <label>IP Address</label>
       <default>*</default>
     </input>
+    <input type="text" token="tkn_port" searchWhenChanged="true">
```
hardikhdholariya, line 33: Should it be the separate filter for src port and dest port?

VatsalJagani: For now because there is only single IP filter we are keeping single port filter.
If there is a need to separate in the future, we'll do that.

hardikhdholariya (approved):

PR #595, opened 2024-07-02
CY-646: password expired alert

mahirchavda:


cyences_app_for_splunk/default/savedsearches.conf
```diff
@@ -4645,7 +4645,7 @@ action.cyences_send_email_action = 1
 action.cyences_notable_event_action.products = Windows, Windows AD
 
 
-[AD - Login Failure due to Password Expired]
+[AD - Login Failure Due To Password Expired]
```
VatsalJagani, line 4648: Alert name change will have some consequences. FYA @hardikhdholariya

VatsalJagani (approved):

cyences_app_for_splunk/default/savedsearches.conf
hardikhdholariya: I'll change alert name on docs.

hardikhdholariya (approved):

PR #606, opened 2024-07-30
Use public ip for cisco ise vpn failure log

mahirchavda:


cyences_app_for_splunk/default/props.conf
```diff
@@ -225,7 +225,7 @@ EVAL-dest_category = if(subtype="vpn" AND vendor_action IN ("tunnel-up", "phase2
 ### Cisco VPN ###
 #################
 [cisco:ise:syslog]
-EVAL-dest_category = if(MESSAGE_CODE IN (5400, 5401) AND isnotnull(AD_User_Candidate_Identities), "vpn_auth", dest_category)
+EVAL-dest_category = if(MESSAGE_CODE IN (5400, 5401) AND isnull(AD_User_Candidate_Identities), "vpn_auth", dest_category)
```
VatsalJagani, line 228: Is this expected? Was this a bug?

mahirchavda: The previous condition was giving event with private ip. Updated condition is for event with public ip

VatsalJagani (approved):

PR #611, opened 2024-08-08
UI for team email configuration

hardikhdholariya:


cyences_app_for_splunk/default/macros.conf
```diff
@@ -1016,6 +1016,14 @@ definition = 10
 definition = ("cyences.com")
 # Add the comma separated values inside the brackets.
 
+# Team Email Configuration
+[cs_soc_team_email]
+definition = soc@crossrealms.com
```
VatsalJagani, line 1021: We need to start with empty somehow.

cyences_app_for_splunk/default/macros.conf
```diff
@@ -1016,6 +1016,14 @@ definition = 10
...
+# Put it with double quote on both side (compulsory)
+
+[cs_compliance_team_email]
+definition = compliance@crossrealms.com
```
VatsalJagani, line 1025: We need to start with empty somehow.

VatsalJagani (changes requested):

cyences_app_for_splunk/default/macros.conf
```diff
@@ -1016,6 +1016,12 @@ definition = 10
 definition = ("cyences.com")
 # Add the comma separated values inside the brackets.
 
+# Team Email Configuration
+[cs_soc_team_email]
+definition =
```
VatsalJagani, line 1021: Are we handling what to do when this macro is empty?

cyences_app_for_splunk/default/macros.conf
```diff
@@ -1016,6 +1016,12 @@ definition = 10
 definition = ("cyences.com")
 # Add the comma separated values inside the brackets.
 
+# Team Email Configuration
+[cs_soc_team_email]
+definition = 
+
+[cs_compliance_team_email]
+definition =
```
VatsalJagani, line 1024: Are we handling what to do when this macro is empty?

VatsalJagani (approved):

cyences_app_for_splunk/default/macros.conf
hardikhdholariya: Yes, in case of empty definition, SOC alerts will go to only default receipients if it's configured otherwise no one get an alert.

cyences_app_for_splunk/default/macros.conf
hardikhdholariya: Yes, in case of empty definition, Compliance alerts will go to only default receipients if it's configured otherwise no one get an alert.

hardikhdholariya:
Created new PR to merge it to 5.0.0-changes branch

PR #615, opened 2024-08-20
remove the UI from the akert page

hardikhdholariya:


hardikhdholariya:
For this change, I'll add upgrade steps to copy the config from savedsearches,conf to alert_actions.conf so it should work as is and display on cyences setup page.

mahirchavda (approved):

VatsalJagani (approved):

PR #617, opened 2024-08-27
removed macro dependency from other apps

hardikhdholariya:


cyences_app_for_splunk/default/macros.conf
```diff
@@ -339,6 +339,14 @@ iseval = 0
 definition = ((`cs_tenable_assets`) OR (`cs_nessus_assets`) OR (`cs_qualys_hostsummary`))
 iseval = 0
 
+[cs_cim_authentication_indexes]
```
VatsalJagani, line 342: Shall we just use cs_authentication_indexes directly??

VatsalJagani (changes requested):

cyences_app_for_splunk/default/macros.conf
hardikhdholariya: I think we have to keep separate macro for the datamodel base query as we are using the cs_authentication_indexes macro and other o365 source macro in tstats searches as shown in SS.

I'll update the definition to index=* instead of cs_authentication_indexes macro as we are adding filter in the tstat searches.

image

VatsalJagani (approved):

PR #621, opened 2024-09-02
Added flow_map_viz custom vizualization

hardikhdholariya:


VatsalJagani (changes requested):
Are we copy pasting it from some app, if that is the case, let's keep it as dependency for now.

hardikhdholariya:
> Are we copy pasting it from some app, if that is the case, let's keep it as dependency for now.

Yes, it's copied from flow_map_viz app. I'm removing it.

VatsalJagani (approved):

PR #626, opened 2024-09-05
Added app dependencies table for each products

hardikhdholariya:


cyences_app_for_splunk/bin/cs_product_list.py
```diff
@@ -79,6 +120,16 @@ def build_source_reviewer_search(by, values, first_call=True):
     },
     {
         "name": "MSSQL",
+        "app_dependencies": [
+            {
+                "label": "Splunk DB Connect",
```
VatsalJagani, line 125: I'm not sure about this. As this will be HF.

VatsalJagani (changes requested):

VatsalJagani (approved):

PR #627, opened 2024-09-06
updated outbound traffic alert

hardikhdholariya:


cyences_app_for_splunk/default/macros.conf
```diff
@@ -383,12 +383,12 @@ definition = 10
...
 [cs_network_ddos_min_m_packets]
-# no. of buckets in millions
+# no. of packets in millions
 definition = 0.5
 
 [cs_network_outbound_min_m_packets]
-# no. of buckets in millions
-definition = 0.5
+# no. of packets in millions
+definition = <TODO>
```
VatsalJagani, line 391: What should be the value here?

VatsalJagani (changes requested):

cyences_app_for_splunk/default/macros.conf
hardikhdholariya: On cerberos, out of 6150 src_ips 45 src_ips have more than 1 million packet transfer for outbound. May be we can set it to 1M.

image

VatsalJagani: If this is hourly then 1 million bottom limit looks okay

VatsalJagani (approved):

PR #629, opened 2024-09-12
Add sophos firewall alerts

hardikhdholariya:


VatsalJagani (commented):
Don't we need to do any change to product setup python file of setup page??

hardikhdholariya:
> Don't we need to do any change to product setup python file of setup page??

Yes, raised separate PR for that to merge it in master.

VatsalJagani (approved):

PR #636, opened 2024-09-18
Moved the splunklib to the lib folder

hardikhdholariya:


VatsalJagani (changes requested):
Need to update github action file with **splunk_python_sdk_install_path** parameter

mahirchavda (approved):

.github/workflows/on_master.yml
```diff
@@ -24,6 +24,7 @@ jobs:
           my_github_token: ${{ secrets.MY_GITHUB_TOKEN }}
           logger_log_files_prefix: "cyences"
           logger_sourcetype: "cyences:logs"
+          splunk_python_sdk_install_path: "lib"
```
VatsalJagani, line 27: Use v4.1 instead of v4

VatsalJagani (changes requested):

VatsalJagani (approved):

PR #637, opened 2024-09-19
removed everything related to honeydb and blocked IP

hardikhdholariya:


cyences_app_for_splunk/default/macros.conf
```diff
@@ -1088,30 +1082,10 @@ iseval = 0
 definition = search *
 iseval = 0
 
-[cs_palo_ddos_prevented_filter]
```
VatsalJagani, line 1091: This is for alert, are we removing this alert as well?

VatsalJagani (changes requested):
Just 1 comment, please check,

cyences_app_for_splunk/default/macros.conf
hardikhdholariya: This is used for the reports.

VatsalJagani (approved):

PR #644, opened 2024-09-26
Update unusual outbound traffic alert

hardikhdholariya:


cyences_app_for_splunk/default/macros.conf
```diff
@@ -386,6 +386,9 @@ definition = 0.5
 # no. of packets in millions
 definition = 0.5
 
+[cs_network_outbound_min_MB_traffic]
+definition = 10
```
VatsalJagani, line 390: Not sure on this value

hardikhdholariya: As outbound traffic varies for each src_ip maybe we can keep the minimal value 1 MB to avoid the false positives.

VatsalJagani: The problem is what if someone sending empty packets to attack the firewall somehow?

hardikhdholariya: yes, correct but this is outbound traffic so I guess this is not a case of attacker, right?

actually upperbound value becomes 0 due to very low traffic. Can we use the max value of the traffic instead of avg. traffic only when avg. traffic is 0? this way we can see some valid upperbound instead 0.

cyences_app_for_splunk/default/savedsearches.conf
```diff
@@ -3040,7 +3040,7 @@ search = | tstats `cs_summariesonly_network_traffic` sum(All_Traffic.bytes) as t
     | fields - total_m_packets_] \
 | stats sum(total_m_packets) as total_m_packets, sum(total_MB) as total_MB, list(dest_ip_mix) as dest_ip_mix, list(action_mix) as actions by src_ip \
 | eval top5_dest_ip=mvindex(dest_ip_mix,0,4) | fields - dest_ip_mix \
-| where total_m_packets>`cs_network_outbound_min_m_packets` \
+| where total_m_packets>`cs_network_outbound_min_m_packets` OR total_MB> `cs_network_outbound_min_MB_traffic` \
```
VatsalJagani, line 3043: For or condition min MB is 10 GB equivalent

VatsalJagani (approved):

PR #646, opened 2024-09-30
upgrade steps for the alert renaming

hardikhdholariya:


cyences_app_for_splunk/bin/upgrade_steps.py
```diff
@@ -118,21 +120,151 @@ def upgrade_5_0_0(session_key, logger):
...
+                ),
+                postargs={"data": json.dumps(payload)},
+                method="POST",
+                sessionKey=session_key,
+                raiseAllErrors=True,
+            )
+
+            logger.info("Re-enabled the product={}".format(product))
+    except Exception:
+        logger.info("Error while product enable/disable. Please manually disable/enable the product on Cyences App Configuration > Product Setup page")
```
VatsalJagani, line 160: Add a warning or error message instead
Also add the exception message

cyences_app_for_splunk/bin/upgrade_steps.py
```diff
@@ -118,21 +120,151 @@ def upgrade_5_0_0(session_key, logger):
...
+
+    alerts_to_disable = [
```
VatsalJagani, line 162: Add comment on why we are doing this:

Removed alert might be present in apps/cyences_app_for_splunk/local/savedsearches.conf, which would constantly generate errors or run searches in all times, etc, etc. To avoid this we are disabling all these alerts if they present in the local folder.

VatsalJagani (changes requested):

VatsalJagani (approved):

PR #648, opened 2024-10-03
Show all event in the window patch dashboard

mahirchavda:


cyences_app_for_splunk/default/data/ui/views/cs_windows_patch.xml
```diff
@@ -73,14 +73,8 @@
         <search>
           <query>`cs_wineventlog_system` host="$host$" $tkn_host_filter$ $tkn_patch_numbers$ EventCode=19 SourceName="Microsoft-Windows-WindowsUpdateClient" NOT "Microsoft Endpoint Protection" NOT "Microsoft Defender Antivirus" 
 | eval message=replace(Message, "^Installation Successful: Windows successfully installed the following update: ", "") 
-| append
```
VatsalJagani, line 76: We used to have this to detect which of the machine don't have any patching events. (lookup)

cyences_app_for_splunk/default/data/ui/views/cs_windows_patch.xml
```diff
@@ -100,14 +94,8 @@
         <search>
           <query>`cs_wineventlog_system` host="$host$" $tkn_host_filter$ $tkn_patch_numbers$ EventCode=19 ("Microsoft Endpoint Protection" OR "Microsoft Defender Antivirus") 
 | eval message=replace(Message, "^Installation Successful: Windows successfully installed the following update: ", "") 
-| append 
-    [| inputlookup all_windows_hosts.csv 
-    | search host="$host$" $tkn_host_filter$ 
-    | eval no_event="No Update Event Found"]
```
hardikhdholariya, line 106: This part will add additional info for the host which does not have any update event.

cyences_app_for_splunk/default/data/ui/views/cs_windows_patch.xml
VatsalJagani: We can use eventstats command (inside appendpipe) to check if there is any events from host is present or not. And if not present then add additional events.

```
| eval event_from="event"
| append [| inputlookup <lookup-name> | eval event_from="lookup"]
| appendpipe [| eventstats count(eval(if(event_from="event", 1, null()))) as no_of_events_from_host by host 
| eval no_event="No Update Event Found"
| where no_of_events_from_host=0]
```

mahirchavda: I think creating separate dashboard for history would be better. The existing one was just doing that.

VatsalJagani (approved):

hardikhdholariya (approved):

PR #651, opened 2024-10-04
Updated the product list with new categories

hardikhdholariya:


cyences_app_for_splunk/bin/cs_product_list.py
```diff
@@ -468,6 +472,23 @@ def build_app_dependency_search(app_list):
             }
         ],
     },
+    {
+        "name": "Vulnerability",
+        "app_dependencies": [],
+        "metadata_count_search": '`cs_vulnerabilities_indexes` sourcetype="*vuln*" | stats count',
```
VatsalJagani, line 478: Sourcetype may not contain *vuln*
We need to stick with what we are using in the datamodel, alert or dashboard.

cyences_app_for_splunk/bin/cs_product_list.py
```diff
@@ -323,6 +287,46 @@ def build_app_dependency_search(app_list):
...
+                "earliest_time": "-7d@d",
+                "latest_time": "now",
+            }
+        ],
+    },
+    {
+        "name": "Network",
+        "app_dependencies": [],
+        "metadata_count_search": '`cs_network_indexes` tag=network tag=communicate | stats count',
+        "macro_configurations": [
```
VatsalJagani, line 317: This stuff would be very useful

cyences_app_for_splunk/default/macros.conf
```diff
@@ -1007,6 +1007,11 @@ definition = NOT activityDisplayName IN ("ApprovalNotification_Create", "Feature
 iseval = 0
 
 # Email
+[cs_email_indexes]
+definition = `cs_o365` OR `cs_gws`
```
VatsalJagani, line 1011: Let's use this macro everywhere related to email.

cyences_app_for_splunk/default/macros.conf
```diff
@@ -1040,6 +1045,11 @@ iseval = 0
 definition = search *
 iseval = 0
 
+[cs_network_indexes]
```
VatsalJagani, line 1048: Let's use this macro everywhere we are using network related stuff.

VatsalJagani (changes requested):

cyences_app_for_splunk/default/macros.conf
hardikhdholariya: For network, we are using "Network Traffic" datamodel and index used from CIM app `cim_Network_Traffic_indexes`

Can we use it directly?

VatsalJagani: Yes, but we are not using keywork **cim** anymore right??

hardikhdholariya: yes but the network traffic model we are using from CIM app so all alerts internally using the `cim_Network_Traffic_indexes` macro.

VatsalJagani: ohh okay, that macro is not present in Cyences then? If not present then it is not recommend to change from our app

VatsalJagani (approved):

PR #660, opened 2024-10-14
Create alert for the multiple user login from the same ip

hardikhdholariya:


cyences_app_for_splunk/default/savedsearches.conf
```diff
@@ -2834,6 +2834,58 @@ action.cyences_notable_event_action.products = Office 365
 action.cyences_notable_event_action.teams = SOC
 
 
+[O365 - Successful Login by Multiple Users from Same IP]
```
VatsalJagani, line 2837: **for** Multiple Users

VatsalJagani (changes requested):

VatsalJagani (approved):

PR #662, opened 2024-10-15
Fix macro name as per changes in dependent app

hardikhdholariya:


VatsalJagani (commented):
Don't we need to add the macro in macros.conf?

hardikhdholariya:
> Don't we need to add the macro in macros.conf?

This macro is from content update app. and that app has changed its name

PR #671, opened 2024-10-21
Upgrade steps changes

hardikhdholariya:


cyences_app_for_splunk/bin/upgrade_steps.py
```diff
@@ -113,39 +113,39 @@ def upgrade_5_0_0(session_key, logger):
     logger.info("Updated the {} macro with the default emails configured for the cyences_send_email_action.".format(SOC_EMAIL_CONFIG_MACRO))
 
     vpn_login_attempt_old_macro_name = "cs_authentication_vpn_login_attemps_outside_working_hour_filter"
```
VatsalJagani, line 115: can we follow the variable naming convention here as well.

cyences_app_for_splunk/bin/upgrade_steps.py
```diff
@@ -113,39 +113,39 @@ def upgrade_5_0_0(session_key, logger):
...
-    vpn_login_attempt_new_macro_name = "cs_authentication_vpn_login_attempts_outside_working_hour_filter"
+    new_alert_name = "Authentication - VPN Login Attempts Outside Working Hours"
     try:
```
VatsalJagani, line 117: Can we move this block of code into a separate util function, we have same code for multiple macro and alert?

VatsalJagani (changes requested):

cyences_app_for_splunk/bin/upgrade_steps.py
hardikhdholariya: Yes, this will require for upgrade steps only so creating function in same file.

VatsalJagani (approved):

PR #683, opened 2024-11-08
App version and release notes

hardikhdholariya:


cyences_app_for_splunk/default/app.conf
```diff
@@ -12,7 +12,7 @@ label = Cyences App for Splunk
 [launcher]
 author = CrossRealms International
 description = Cyences App for Splunk built by CrossRealms International
-version = 5.0.0
+version = 5.1.0
```
VatsalJagani, line 15: Should we do 5.0.1?

docs/release_notes/old_release_notes.md
```diff
@@ -8,6 +8,289 @@ parent: Release Notes
 
 
 
+## Version 5.0.0 (October 2024)
```
VatsalJagani, line 11: Let's keep this to main release notes if we are doing v.5.0.1

VatsalJagani (changes requested):

cyences_app_for_splunk/default/app.conf
hardikhdholariya: Yes, 5.0.1 as it's minor release.

VatsalJagani (approved):

PR #688, opened 2024-11-28
is_public_ip and is_private_ip macros added.

VatsalJagani:


cyences_app_for_splunk/default/macros.conf
```diff
@@ -94,6 +94,22 @@ args = ip_field
 definition = eval is_source_internal = if(cidrmatch("10.0.0.0/8",$ip_field$) OR cidrmatch("172.16.0.0/12",$ip_field$) OR cidrmatch("192.168.0.0/16",$ip_field$), "Yes", "No") 
 iseval = 0
 
+[cs_is_private_ip(1)]
```
hardikhdholariya, line 97: we have similar macro for this defination: **cs_identify_private_ips**

cyences_app_for_splunk/default/macros.conf
```diff
@@ -94,6 +94,22 @@ args = ip_field
...
+args = ip_field
+definition = eval is_private_ip = if(cidrmatch("10.0.0.0/8",$ip_field$) OR cidrmatch("172.16.0.0/12",$ip_field$) OR cidrmatch("192.168.0.0/16",$ip_field$), "Yes", "No") 
+iseval = 0
+
+[cs_is_public_ip(1)]
+args = ip_field
+definition = eval is_public_ip = if( \
+    (match($ip_field$, "^(?:\d{1,3}\.){3}\d{1,3}$") OR \
+    match($ip_field$, "^([0-9a-fA-F]{1,4}:){7}([0-9a-fA-F]{1,4}|\*)$|^([0-9a-fA-F]{1,4}:){1,7}:$|^::([0-9a-fA-F]{1,4}:){0,6}[0-9a-fA-F]{1,4}$")) AND \
```
hardikhdholariya, line 106: Do we need to validate the format?

hardikhdholariya: are we using these macros anywhere? or just for the future use?

cyences_app_for_splunk/default/macros.conf
VatsalJagani: But new macro name seems more appropreaite.

Existing macro might be in use already somewhere at customer filters, hence I have not renamed and created new one.

cyences_app_for_splunk/default/macros.conf
VatsalJagani: For use in filter macros.
For example, increase severity if ip is public ip, etc.

cyences_app_for_splunk/default/macros.conf
hardikhdholariya: This macro is not available to modify from the cyences UI as this is the common definition. You can remove cs_identify_private_ips and replace the name cs_is_private_ip everywhere.

hardikhdholariya (approved):

PR #705, opened 2025-01-17
Add alerts for the cisco ios

hardikhdholariya:


cyences_app_for_splunk/default/savedsearches.conf
```diff
@@ -3354,6 +3354,41 @@ action.cyences_notable_event_action.products = Cisco IOS
 action.cyences_notable_event_action.teams = SOC
 
 
+[Cisco IOS - CPU Utilization Exceeds the Threshold]
```
mahirchavda, line 3357: I feel this is not security alert. rather performance monitoring

mahirchavda (approved):

cyences_app_for_splunk/default/savedsearches.conf
hardikhdholariya: Yes, it's on performance side. I confirmed with compliance team and it's helpful for them so I'm moving it to compliance.

PR #706, opened 2025-01-31
fortigate vpn auth event config changes

hardikhdholariya:


cyences_app_for_splunk/default/tags.conf
```diff
@@ -39,19 +39,17 @@ end = enabled
 #####################
 ### Fortigate VPN ###
 #####################
-[eventtype=cs_fortigate_vpn_login]
+[eventtype=cs_fortigate_vpn_auth]
 authentication = enabled
 vpn = enabled
-network = enabled
```
VatsalJagani, line 45: Don't we need this tags?

cyences_app_for_splunk/default/props.conf
```diff
@@ -203,18 +203,19 @@ EXTRACT-src_ip_for_radius = \sfrom \'(?<src_ip>[^\']+)\',
...
 # phase2-up - 
+# negotiate - SSL VPN login success
 # ssl-login-fail - SSL VPN login fail
 # tunnel-down - SSL VPN tunnel down
 # ssl-web-close - SSL VPN close
 
 [fgt_event]
 # fgt_event for old Fortigate Add-on (1.6.2)
-EVAL-dest_category = if(subtype="vpn" AND vendor_action IN ("tunnel-up", "phase2-up", "ssl-login-fail"), "vpn_auth", dest_category)
+EVAL-dest_category = if(subtype="vpn" AND vendor_action IN ("negotiate", "ssl-login-fail"), "vpn_auth", dest_category)
```
VatsalJagani, line 213: Do we negotiate events from any environment?

hardikhdholariya: Yes, we have negotiate events which is VPN login success events.

VatsalJagani (approved):

cyences_app_for_splunk/default/tags.conf
hardikhdholariya: removed tags refers to the network session data model which is not applied to vpn auth events.

I've added network session tags to below eventtypes which has tag start and end as well for session.

PR #712, opened 2025-02-06
fortigate vpn auth logs config changes

hardikhdholariya:


cyences_app_for_splunk/default/props.conf
```diff
@@ -210,12 +210,12 @@ EXTRACT-src_ip_for_radius = \sfrom \'(?<src_ip>[^\']+)\',
 
 [fgt_event]
 # fgt_event for old Fortigate Add-on (1.6.2)
-EVAL-dest_category = if(subtype="vpn" AND vendor_action IN ("negotiate", "ssl-login-fail"), "vpn_auth", dest_category)
+EVAL-dest_category = if(subtype="vpn" AND vendor_action IN ("tunnel-up", "ssl-login-fail"), "vpn_auth", dest_category)
```
VatsalJagani, line 213: Just tunnel-up won't work, because tunnel-up also includes Network tunnels, which is not expected.

VatsalJagani (changes requested):

cyences_app_for_splunk/default/eventtypes.conf
```diff
@@ -31,7 +31,7 @@ search = sourcetype="pan:system" object="RSA_Radius" dest_category="radius_auth"
 ### Fortigate VPN ###
 #####################
 [cs_fortigate_vpn_auth]
-search = sourcetype IN ("fgt_event", "fortigate_event") subtype="vpn" vendor_action IN ("negotiate", "ssl-login-fail")
+search = sourcetype IN ("fgt_event", "fortigate_event") subtype="vpn" vendor_action IN ("tunnel-up", "ssl-login-fail")
```
VatsalJagani, line 34: Need the change here as well.

cyences_app_for_splunk/default/props.conf
```diff
@@ -210,12 +210,12 @@ EXTRACT-src_ip_for_radius = \sfrom \'(?<src_ip>[^\']+)\',
...
-EVAL-dest_category = if(subtype="vpn" AND vendor_action IN ("negotiate", "ssl-login-fail"), "vpn_auth", dest_category)
+EVAL-dest_category = if(subtype="vpn" AND ((vendor_action="tunnel-up" reason="login successfully") OR vendor_action="ssl-login-fail"), "vpn_auth", dest_category)
 
 
 [fortigate_event]
 # fortigate_event for the new Fortigate Add-on (1.6.5)
-EVAL-dest_category = if(subtype="vpn" AND vendor_action IN ("negotiate", "ssl-login-fail"), "vpn_auth", dest_category)
+EVAL-dest_category = if(subtype="vpn" AND ((vendor_action="tunnel-up" reason="login successfully") OR vendor_action="ssl-login-fail"), "vpn_auth", dest_category)
 # Note - eval action is not working because Fortigate Add-on's lookup is overriding the action field, hence we need to make changes in the data-model definition
 # action field update in the Authentication data-model -> case(sourcetype="fgt_event" AND subtype="vpn" AND vendor_action IN ("tunnel-up", "phase2-up"), "success", sourcetype="fgt_event" AND subtype="vpn" AND vendor_action="ssl-login-fail", "failure", isnull(action) OR action="", "unknown", 1==1, action)
```
VatsalJagani, line 220: Do we need to change this as per new logic for action field - failure & success?

VatsalJagani (changes requested):

cyences_app_for_splunk/default/props.conf
hardikhdholariya: as per note, action field is overriding from fortigate addon. so I'm testing it and after will change.

cyences_app_for_splunk/default/data/ui/views/cs_vpn_reports.xml
```diff
@@ -231,7 +232,7 @@
       <table>
         <title>Successful Session</title>
         <search>
-          <query>`cs_vpn_indexes` dest_category="vpn_auth" action="success" user IN $User$ dest=$tkn_dest$ src=$tkn_public_ip$ 
+          <query>`cs_vpn_indexes` dest_category="vpn_auth" action="added" user IN $User$ dest=$tkn_dest$ src=$tkn_public_ip$
```
VatsalJagani, line 235: We need to put only specific tunnel-down, just like tunnel-up

cyences_app_for_splunk/default/props.conf
VatsalJagani: Let's write a note here that fortigate Add-on is overriding the action field, hence we are not writing the props config here and instead writing it in the data-model & dashboard queries.

VatsalJagani (approved):

PR #715, opened 2025-02-12
removed unnecessary source

hardikhdholariya:


cyences_app_for_splunk/default/macros.conf
```diff
@@ -1593,9 +1593,6 @@ definition = search *
 [cs_aws]
 definition = index IN (aws,summary)
 
-[cs_aws_login_filter_source]
-definition = (sourcetype="aws:cloudtrail" Authentication.signature="ConsoleLogin")
```
VatsalJagani, line 1597: @hardikhdholariya  - I'm guessing this is already has tag=authentication & action field.

VatsalJagani (approved):

PR #720, opened 2025-02-19
Release notes for 520

hardikhdholariya:


docs/release_notes/old_release_notes.md
```diff
@@ -7,6 +7,54 @@ parent: Release Notes
 ---
 
 
+## Version 5.1.0 (December 2024)
+
+* ### SOC AI Integration
```
VatsalJagani, line 12: This should be in the Documentation at two place:
* User Doc -> What is this feature & how is it useful & how it is being used.
* Configuration -> How can you configure it. (Mention contacting Crossrealms - Need to confirm that with Usama I think on what contact stuff to put)

VatsalJagani: I know this is for last version but let's make sure this is added.

hardikhdholariya: Yes, we already have at configuration section & I'll update the contact info once I have. 
on user doc, I'll add.

VatsalJagani (approved):

PR #726, opened 2025-03-03
removed the service principal logins

hardikhdholariya:


cyences_app_for_splunk/default/data/ui/views/cs_o365_reports.xml
```diff
@@ -176,7 +176,7 @@
       <title>Login by Location</title>
       <map>
         <search>
-          <query>`cs_o365` sourcetype="o365:management:activity" Workload=AzureActiveDirectory Operation IN("UserLoggedIn", "UserLoginFailed") $tkn_host$ 
+          <query>`cs_o365` sourcetype="o365:management:activity" Workload=AzureActiveDirectory Operation IN("UserLoggedIn", "UserLoginFailed") user!="not available" user_type!="ServicePrincipal" $tkn_host$
```
VatsalJagani, line 179: Is the user_type field present in 100% of the events for below search??

* `cs_o365` sourcetype="o365:management:activity" Workload=AzureActiveDirectory Operation IN("UserLoggedIn", "UserLoginFailed")

hardikhdholariya: We are getting **UserType** in 100% of the event but **user_type** (which is mapped from the o365 TA props.conf) is in 99.5% of the events. 

& missing user_type field is for the UserType=9 for that type we are getting empty clientIP and user as a GUID format. It seems useful but required more research as sufficient details are not there. 
https://learn.microsoft.com/en-us/purview/audit-log-detailed-properties#usertype-and-userkey-scenarios

I'll add fillnull in props so we can't miss it.

VatsalJagani (approved):

PR #728, opened 2025-03-07
Added report to assign privilege to domain admins

hardikhdholariya:


cyences_app_for_splunk/default/savedsearches.conf
```diff
@@ -6973,6 +6973,36 @@ search =  index=* NOT sourcetype IN ("meraki", "fortigate_utm", "WindowsUpdateLo
 | cyencesusermanager operation="addentries" 
 
 
+[User Inventory - Assign Privileges To AD Domain Admins]
+disabled = 0
+enableSched = 1
+alert.track = 0
+cron_schedule = 0 1 * * *
```
VatsalJagani, line 6980: This doesn't need to be run every hour. Maybe once a day or even once a week is fine.

VatsalJagani (changes requested):

cyences_app_for_splunk/default/savedsearches.conf
hardikhdholariya: It's running every day at 1 AM.

PR #734, opened 2025-03-21
updated the alert

hardikhdholariya:


cyences_app_for_splunk/default/savedsearches.conf
```diff
@@ -6162,17 +6162,19 @@ display.page.search.tab = statistics
 display.page.search.mode = fast
 request.ui_dispatch_app = cyences_app_for_splunk
 request.ui_dispatch_view = search
-search = `cs_linux_auditd` type=SERVICE_STOP unit IN ("firewalld", "ufw") \
+search = `cs_linux_auditd` type IN (SERVICE_START, SERVICE_STOP) unit IN ("firewalld", "ufw") \
+| dedup host \
+| search type="SERVICE_STOP" \
```
crmpandya, line 6167: If only SERVICE_STOP type events are required, why are SERVICE_START type events being added at the beginning?

crmpandya (changes requested):

cyences_app_for_splunk/default/savedsearches.conf
hardikhdholariya: due to system reboot, service will stop and start again so in that case, we'll consider latest status of that host and not trigger alert if it already started.

cyences_app_for_splunk/default/savedsearches.conf
```diff
@@ -6162,17 +6162,19 @@ display.page.search.tab = statistics
...
 | table _time host type UID comm exe unit \
 | rename comm as command \
 | `cs_human_readable_time_format(_time, event_time)` \
 | eval cyences_severity = "high" \
 | `cs_linux_system_firewall_service_stopped_filter`
 action.cyences_notable_event_action = 1
 action.cyences_notable_event_action.param.filter_macro_name = cs_linux_system_firewall_service_stopped_filter
-action.cyences_notable_event_action.contributing_events =  `cs_linux_auditd` type=SERVICE_STOP unit IN ("firewalld", "ufw")
+action.cyences_notable_event_action.contributing_events =  `cs_linux_auditd` type IN (SERVICE_START, SERVICE_STOP) unit IN ("firewalld", "ufw") | dedup host | search type="SERVICE_STOP"
```
VatsalJagani, line 6175: Service start event I think should have less severity than service stop.

VatsalJagani (changes requested):

crmpandya (approved):

cyences_app_for_splunk/default/savedsearches.conf
hardikhdholariya: alert is only for service_stop. We included both to avoid triggering alert if service restarted again.

PR #744, opened 2025-05-02
Updated ad password change outside working hours alert time and updated kaspersky doc.

crmpandya:


cyences_app_for_splunk/default/savedsearches.conf
```diff
@@ -4732,7 +4732,7 @@ alert.suppress = 0
 counttype = number of events
 quantity = 0
 relation = greater than
-cron_schedule = 0 10 * * *
+cron_schedule = 2-59/30 * * * *
```
hardikhdholariya, line 4735: Need to update earliest and latest time accordingly otherwise it triggers multiple time for same event.

Also, during working hours alert won't trigger so need to check the how to set up schedule in best way

crmpandya: during working hours it won't trigger as there is condition in alert to check it.

PR #751, opened 2025-05-06
Added palo alto dashboard panel.

crmpandya:


cyences_app_for_splunk/default/data/ui/views/cs_paloalto_firewall_reports.xml
```diff
@@ -73,6 +73,29 @@
       </html>
     </panel>
   </row>
+  <row>
+    <panel>
+      <title>Palo Alto Configuration Changes</title>
+      <table>
+        <search>
+          <query>`cs_palo` sourcetype=pan:log Action!="show"
```
VatsalJagani, line 81: @crmpandya , @hardikhdholariya  - Need to validate the query and make sure it include all the configuration changes.

crmpandya: Confirmed with jeeda. it includes all the configuration changes. 

Also discussed with hardik as well. He suggested to use pan:config source instead of pan:log. so accordingly updated the query.

hardikhdholariya (approved):

PR #756, opened 2025-05-22
Updated app version.

crmpandya:


VatsalJagani (changes requested):
We need to update the App Configuration React Page version as well, as there might be changes there as well.

hardikhdholariya (approved):

VatsalJagani (approved):

PR #770, opened 2025-07-16
Added method to read .sqlaudit files from db connect

hardikhdholariya:


docs/data_onboarding/database/mssql.md
```diff
@@ -24,18 +24,30 @@ The Cyences App has support following data collection mechanisms
 1. Window Application log using Splunk Add-on for Microsoft Windows (Recommended)
     * [https://www.ultimatewindowssecurity.com/sqlserver/auditlog/auditobject.aspx](https://www.ultimatewindowssecurity.com/sqlserver/auditlog/auditobject.aspx)
 
-2. Audit table using Splunk DB Connect app 
+2. By reading **Audit table** or reading ***.sqlaudit** files using Splunk DB Connect app
```
VatsalJagani, line 27: This should be 3rd method.

As 3nd method we already used to read data from DB tables.

VatsalJagani: Please keep both method separate.

VatsalJagani (changes requested):

docs/data_onboarding/database/mssql.md
hardikhdholariya: I kept it under same section as most of the steps are similar just query to read table and files are different.

docs/data_onboarding/database/mssql.md
```diff
@@ -24,18 +24,30 @@ The Cyences App has support following data collection mechanisms
...
+
+    **NOTE:** While creating input on DB connect app, create appropriate MSSQL query to read the audit table. If you are reading *.sqlaudit file then use the following MSSQL search:
+
+    ```
+    SELECT *
+    FROM sys.fn_get_audit_file(
+    '<<FILE_PATH>>',
+    DEFAULT,
+    DEFAULT
+    )
```
mahirchavda, line 48: Need to add where condition for time filter to only collect latest data

VatsalJagani: If its doable it would be huge.

hardikhdholariya: directly not possible as user has to go through UI part and look at the right fields. I'll add the input creation steps, it'll be easy to follow

PR #773, opened 2025-07-22
Add filter to ad group membership changes alert

crmpandya:
Screenshot 2025-07-22 at 8 19 09 PM

### DIFF
--- a/cyences_app_for_splunk/default/data/ui/views/cs_ad_reports.xml
+++ b/cyences_app_for_splunk/default/data/ui/views/cs_ad_reports.xml
@@ -75,6 +75,7 @@
           <query>`ms_obj_changes_base_cat_customised("groupmembershipchanged")`
 | search $tkn_search_common|s$ $tkn_search_group_membership_changed|s$
 | `ms_obj_groupmembership_change_out`
+| search NOT group_obj_nm = "NA"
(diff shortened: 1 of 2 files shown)


cyences_app_for_splunk/default/data/ui/views/cs_ad_reports.xml
```diff
@@ -75,6 +75,7 @@
           <query>`ms_obj_changes_base_cat_customised("groupmembershipchanged")`
 | search $tkn_search_common|s$ $tkn_search_group_membership_changed|s$
 | `ms_obj_groupmembership_change_out`
+| search NOT group_obj_nm = "NA"
```
VatsalJagani, line 78: What are these events technically??

crmpandya: When any changes are made to group membership, this event is generated containing the user and group information.

VatsalJagani: I understand but in which case we are receiving these N/A values??

crmpandya: Not sure about scenario. I asked jeeda as well. she was also not sure about scenario.

VatsalJagani (approved):

PR #774, opened 2025-07-23
Fixed user field extraction issue for failed login events for pfsense openvpn logs.

crmpandya:


cyences_app_for_splunk/default/props.conf
```diff
@@ -275,7 +275,9 @@ FIELDALIAS-user = "log.distinguished_name_user" as user
 [pfsense:openvpn]
 EXTRACT-authentication_action = \s(?<authentication_action>[A-Za-z\s]+).?$
 EXTRACT-src = '(?<src>[0-9.]+):\d+'
+EXTRACT-failed_login_user = '(?<failed_login_user>[A-Za-z.@]+)'
```
VatsalJagani, line 278: Nope single quote can be number of things. It needs to be more specific please.

VatsalJagani (approved):

cyences_app_for_splunk/default/props.conf
crmpandya: Updated regex

PR #786, opened 2025-09-02
Added bruteforce attempt for dest alert.

crmpandya:


crmpandya:
> We also need to add similar alert for VPN

Added.

